PR #28, opened 2024-08-09
cleanup: remove `.spec.global` field from the `KongPluginBinding` resource

mlavacca:

This PR removes the `.spec.global` field from the `KongPluginBinding` resource.  KPB is a namespaced resource and having a field in its spec that affects all the resources of a cluster breaks the namespace isolation.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


czeslavo (approved):
Hm, that makes sense. I guess that means that to model global plugins with bindings, we will need a separate `KongPluginClusterBinding` CRD?

mlavacca:
> Hm, that makes sense. I guess that means that to model global plugins with bindings, we will need a separate `KongPluginClusterBinding` CRD?

Yep, I already commented in the original issue about this possibility. I think that's the way to go 👍

PR #33, opened 2024-08-12
chore: always execute the cleanup

mlavacca:

We need to always run the cleanup. In case an error in the creation was occurred, `client.IgnoreNotFound` does not return any error in the creation.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/test/crdsvalidation/kongpluginbindings/kongpluginbindings_test.go
+++ b/test/crdsvalidation/kongpluginbindings/kongpluginbindings_test.go
@@ -29,7 +29,7 @@ func TestKongPluginBindings(t *testing.T) {
 				t.Run(tc.Name, func(t *testing.T) {
 					cl := cl.KongPluginBindings(tc.KongPluginBinding.Namespace)
 					kpb, err := cl.Create(ctx, &tc.KongPluginBinding, metav1.CreateOptions{})
-					if err != nil {
+					if err == nil {
 						t.Cleanup(func() {
 							assert.NoError(t, client.IgnoreNotFound(cl.Delete(ctx, kpb.Name, metav1.DeleteOptions{})))
 						})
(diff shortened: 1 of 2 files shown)


czeslavo (approved):

test/crdsvalidation/kongpluginbindings/kongpluginbindings_test.go
```diff
@@ -29,11 +29,13 @@ func TestKongPluginBindings(t *testing.T) {
 				t.Run(tc.Name, func(t *testing.T) {
 					cl := cl.KongPluginBindings(tc.KongPluginBinding.Namespace)
 					kpb, err := cl.Create(ctx, &tc.KongPluginBinding, metav1.CreateOptions{})
-					if tc.ExpectedErrorMessage == nil {
-						assert.NoError(t, err)
+					if err != nil {
```
pmalek, line 32: If we failed to create the resource then we don't need to schedule the deletion, no?

Can you elaborate why we need to change this? In case there was an error (in the previous version) line 33 will assert and log and then we'll schedule a testing cleanup which will ignore the not found error. Am I missing something?

mlavacca: With the previous logic, if `expectedErrorMessage` != nil (error expected), but the creation succeeded (no error occurred), we perform no cleanup. This way we always perform the cleanup, no matter the expected outcome. Does it make sense?

pmalek (approved):

test/crdsvalidation/kongpluginbindings/kongpluginbindings_test.go
pmalek: Oh I see. Yup, that makes sense. 👍

PR #45, opened 2024-08-16
feat: add ExtractTags

pmalek:

Add `ExtractTags` and related exported consts.

### DIFF
--- a/pkg/metadata/consts.go
+++ b/pkg/metadata/consts.go
@@ -1,0 +1,9 @@
+package metadata
+
+const (
+	// AnnotationPrefix is the prefix for Kong annotations.
+	AnnotationPrefix = "konghq.com"
+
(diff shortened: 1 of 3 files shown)


pkg/metadata/consts.go
```diff
@@ -0,0 +1,9 @@
+package metadata
+
+const (
+	// AnnotationPrefix is the prefix for Kong annotations.
+	AnnotationPrefix = "konghq.com"
+
+	// TagKey is the key for the tags annotation.
+	TagKey = "/tags"
```
czeslavo, line 8: Having the `ExtractTags` use them, to minimize the amount of exported symbols we might leave these private. WDYT?

pmalek: 👍 f59ee6b2263763bde8235c912f3acca57d6564eb

PR #54, opened 2024-08-22
feat: add missing KongConsumerGroup methods

pmalek:

Part of https://github.com/Kong/gateway-operator/issues/514

Required for https://github.com/Kong/gateway-operator/pull/510


api/configuration/v1beta1/kongconsumergroup_types.go
```diff
@@ -28,20 +31,75 @@ import (
...
+type KongConsumerGroupSpec struct {
+	// Name is the name of the ConsumerGroup in Kong.
+	Name *string `json:"name,omitempty"`
+
+	// ControlPlaneRef is a reference to a ControlPlane this ConsumerGroup is associated with.
+	// +optional
+	ControlPlaneRef *configurationv1alpha1.ControlPlaneRef `json:"controlPlaneRef,omitempty"`
+}
+
+func (c *KongConsumerGroup) InitKonnectStatus() {
```
czeslavo, line 57: How about keeping this private? It's used in setters, but I don't think we should leak this outside (the same applies to KongConsumer, KongRoute and KongService).

czeslavo (approved):

PR #57, opened 2024-08-27
chore: fixed default programmed condition

mlavacca:

The default `Unknown` programmed condition has been correctly set for all the configuration CRDs.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


api/configuration/v1/kongconsumer_types.go
```diff
@@ -61,6 +61,7 @@ type KongConsumer struct {
 	Spec KongConsumerSpec `json:"spec,omitempty"`
 
 	// Status represents the current status of the KongConsumer resource.
+	// +kubebuilder:default={conditions: {{type: "Programmed", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
```
czeslavo, line 64: Didn't `+kubebuilder:default` on `Conditions` work? If not, why? 🤔

mlavacca: Because the status is marked as optional, hence it is not initialized, and the `+kubebuilder:default` on `Conditions` is not applied. With this new annotation, we initialize the status and the conditions within.

czeslavo: Ahh, it's the `omitempty` making the status optional effectively. Makes sense now.

czeslavo (approved):

PR #61, opened 2024-09-03
chore: rename KonnectControlPlane to KonnectGatewayControlPlane

czeslavo:

Although the entity name on the Konnect side is "control plane", it will be better to make it explicit as in the future we'll also have a KonnectMeshControlPlane.

**Which issue this PR fixes**

On @mheap's request.


programmer04 (commented):
Sound good, please resolve conflicts @czeslavo

PR #82, opened 2024-09-18
feat(konnect): add namespace in KonnectNamespacedRef

### LINKED ISSUE #36: Allow cross namespaced Konnect Control Plane references
## Problem Statement

[`ControlPlaneRef`](https://github.com/Kong/kubernetes-configuration/blob/d4d87e1acd2d5a7d333c30504c1d0aa97e6b9809/api/configuration/v1alpha1/konnect_controlplaneref_types.go#L8-L9) allows setting the referenced ControlPlane.

One of the types of this reference is `KonnectNamespacedRef`.

This task tracks the implementation of cross namespace references between entities and Konnect ControlPlanes.

## Additional Information

Related #9

## Acceptance Criteria

- [ ] As a user, I can reference a KonnectControlPlane in a different Namespace.

randmonkey:
Add `namespace` in `konnectNamespacedRef` because some of Kong entities like `KongVault` and `KongClusterPlugin` are cluster scoped and they require the namespace for referring to a Konnect gateway controlplane.
**Which issue this PR fixes**

#36

**PR Readiness Checklist**:

### DIFF
--- a/api/configuration/v1/kongconsumer_types.go
+++ b/api/configuration/v1/kongconsumer_types.go
@@ -34,7 +34,6 @@ import (
 // +kubebuilder:printcolumn:name="Programmed",type=string,JSONPath=`.status.conditions[?(@.type=="Programmed")].status`
 // +kubebuilder:validation:XValidation:rule="has(self.username) || has(self.custom_id)", message="Need to provide either username or custom_id"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec.controlPlaneRef) || has(self.spec.controlPlaneRef)", message="controlPlaneRef is required once set"
-//  REVIEW: allow same namespace?
 // +kubebuilder:validation:XValidation:rule="!has(self.spec.controlPlaneRef.konnectNamespacedRef) ? true : !has(self.spec.controlPlaneRef.konnectNamespacedRef.__namespace__)", message="spec.controlPlaneRef cannot specify namespace for namespaced resource"
 // +kubebuilder:validation:XValidation:rule="(!has(self.status) || !self.status.conditions.exists(c, c.type == 'Programmed' && c.status == 'True')) ? true : oldSelf.spec.controlPlaneRef == self.spec.controlPlaneRef", message="spec.controlPlaneRef is immutable when an entity is already Programmed"
 
--- a/api/configuration/v1/kongconsumer_types.go
+++ b/api/configuration/v1/kongconsumer_types.go
@@ -34,6 +34,8 @@ import (
 // +kubebuilder:printcolumn:name="Programmed",type=string,JSONPath=`.status.conditions[?(@.type=="Programmed")].status`
 // +kubebuilder:validation:XValidation:rule="has(self.username) || has(self.custom_id)", message="Need to provide either username or custom_id"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec.controlPlaneRef) || has(self.spec.controlPlaneRef)", message="controlPlaneRef is required once set"
+//  REVIEW: allow same namespace?
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.controlPlaneRef.konnectNamespacedRef) ? true : !has(self.spec.controlPlaneRef.konnectNamespacedRef.__namespace__)", message="spec.controlPlaneRef cannot specify namespace for namespaced resource"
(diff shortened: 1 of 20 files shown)


pmalek (changes requested):
For now we've decided to not implement cross namespace references.

https://github.com/Kong/gateway-operator/issues/524 for example has been put into 'post-GA' part of the epic.

Let's first solve the immediate scope from https://github.com/Kong/gateway-operator/issues/370 (for October's release).

LMKWYT

randmonkey:
> For now we've decided to not implement cross namespace references.
> 
> [Kong/gateway-operator#524](https://github.com/Kong/gateway-operator/issues/524) for example has been put into 'post-GA' part of the epic.
> 
> Let's first solve the immediate scope from [Kong/gateway-operator#370](https://github.com/Kong/gateway-operator/issues/370) (for October's release).
> 
> LMKWYT

`KongVault` is also cluster scoped. If we do not want to add namespace, it should also be moved out of KGO 1.4 scope.

pmalek:
Left a comment about this in https://kongstrong.slack.com/archives/C079GK3KB9R/p1726654138359599.

IMHO we should prevent using cross-namespace refs until we implement proper handling of those. In order to move forward with Vaults, let's review and merge this but at the same time let's add CEL validation rules to other types that will prevent users from using cross namespace ref in non cluster scoped resources.

api/configuration/v1alpha1/konnect_controlplaneref_types.go
```diff
@@ -41,4 +41,8 @@ type KonnectNamespacedRef struct {
 
 	// TODO: Implement cross namespace references:
 	// https://github.com/Kong/kubernetes-configuration/issues/36
+	// Currently only cluster scoped resources (KongVault) are allowed to set `konnectNamespacedRef.namespace`.
+	// Namespace is the namespace where the Konnect Control Plane is in.
+	// +optional
```
pmalek, line 46: Reword the comment to not include the TODO in the resulting CRDs.

```suggestion
	// TODO: Implement cross namespace references:
	// https://github.com/Kong/kubernetes-configuration/issues/36

	// Namespace is the namespace of the Konnect Control Plane.
	// Currently only cluster scoped resources (KongVault) are allowed to set `konnectNamespacedRef.namespace`.
	//
	// +optional
```

api/configuration/v1/kongconsumer_types.go
```diff
@@ -34,6 +34,8 @@ import (
 // +kubebuilder:printcolumn:name="Programmed",type=string,JSONPath=`.status.conditions[?(@.type=="Programmed")].status`
 // +kubebuilder:validation:XValidation:rule="has(self.username) || has(self.custom_id)", message="Need to provide either username or custom_id"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec.controlPlaneRef) || has(self.spec.controlPlaneRef)", message="controlPlaneRef is required once set"
+//  REVIEW: allow same namespace?
```
pmalek, line 37: Yes please, if possible let's add that rule.

api/configuration/v1/kongconsumer_types.go
```diff
@@ -34,6 +34,8 @@ import (
...
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.controlPlaneRef.konnectNamespacedRef) ? true : !has(self.spec.controlPlaneRef.konnectNamespacedRef.__namespace__)", message="spec.controlPlaneRef cannot specify namespace for namespaced resource"
```
pmalek, line 38: Why `__namespace__` and not `namespace`? 🤔

randmonkey: Because `namespace` is a keyword in CEL rules. You HAVE TO use `__namespace__` to refer to fields with name `namespace`, otherwise you will receive a `compilation failed: ERROR: <input>:1:9: undefined field 'namespace'` error in installing the CRD: https://github.com/kubernetes/kubernetes/issues/122777.

pmalek: TIL. 👍

api/configuration/v1/kongconsumer_types.go
randmonkey: Looks like we cannot access `metadata.namespace` in CEL: https://github.com/kubernetes/kubernetes/issues/122163.

pmalek: So from https://github.com/kubernetes/kubernetes/issues/122163#issuecomment-1908351389 it seems that it is doable with a different order of fields in the generated CRD manifest but apparently kubebuilder doesn't generate the CRDs in this way?

Nonetheless, we don't actually need to allow specifying the namespace here. Users would simply not provide one when the referenced entity exists in the same namespace as the referrer.

pmalek (approved):

PR #83, opened 2024-09-18
feat(crd): add CredentialBasicAuth

pmalek:

This PR adds `CredentialBasicAuth` CRD which will be used by KGO.

Resources of this type will be created when `KongConsumer` will use a `Secret` as credential. This will result in `CredentialBasicAuth` being created, constituting a 1:1 mapping with a credential resource in Konnect.

**Which issue this PR fixes**

Related to https://github.com/Kong/gateway-operator/issues/617

**PR Readiness Checklist**:


api/configuration/v1alpha1/credential_basic_auth_types.go
```diff
@@ -0,0 +1,159 @@
...
+	// +kubebuilder:validation:Required
+	SecretRef corev1.LocalObjectReference `json:"secretRef"`
+
+	CredentialBasicAuthAPISpec `json:",inline"`
+}
+
+// CredentialBasicAuthAPISpec defines specification of a BasicAuth credential.
+type CredentialBasicAuthAPISpec struct {
+	// Password is the password for the BasicAuth credential.
+	Password *string `json:"password,omitempty"`
```
randmonkey, line 127: Should it be redacted?

randmonkey (commented):
So a `CredentialBasicAuth` represents a relationship that a `KongConsumer` referencing a `Secret` as a basic auth credential attached to the consumer?

pmalek:
> So a `CredentialBasicAuth` represents a relationship that a `KongConsumer` referencing a `Secret` as a basic auth credential attached to the consumer?

Yes, that's the idea. If more Consumers use the same credential (not sure how realistic use case this is) this will create multiple credential objects for each usage.

api/configuration/v1alpha1/credential_basic_auth_types.go
pmalek: Hm, that's an interesting idea. Looking at https://book.kubebuilder.io/reference/markers I cannot find a code marker that would allow this. Quick google search doesn't yield any results about this feature of CRD fields. 🤔 

---

An alternative is to not copy the config from the Secret but to get the config inside the generic reconciler when Konnect entity is to be created/updated. I'd prefer to avoid that as that would make the generic reconciler be responsible for even more, whereas with the approach from this PR it will not need to apply what's defined in the CRD. 

Plus, with this approach users can clearly see in the CRD what's the expected state in Konnect.

api/configuration/v1alpha1/credential_basic_auth_types.go
```diff
@@ -0,0 +1,159 @@
...
+	// Tags is a list of tags for the BasicAuth credential.
+	Tags []string `json:"tags,omitempty"`
+	// Username is the username for the BasicAuth credential.
+	Username *string `json:"username,omitempty"`
```
czeslavo, line 131: Given that we know this field is not optional, why not use just `string`? Same applies to the password.

randmonkey: And also we can add a `MinLength=1` constraint to make it non empty.

pmalek: Hm, so the API https://kong-platform-api.netlify.app/konnect/control-planes-config/v2/openapi.yaml/#tag/Basic-auth-credentials/operation/create-basic-auth-with-consumer does not seem to require either of those but Konnect UI does require both:

image

I'll make the changes for this.

pmalek: 46384cea11836e799f2dcc7d942f4dc620ebd396

pmalek:
Blocked for now before we get consensus on how this should look like: 

https://kongstrong.slack.com/archives/C079GK3KB9R/p1726735862213059

pmalek:
Unblocked.

This for now will contain only the consumer ref and verbatim credential field.

This will replace the "old" way of binding consumers to credentials (`Secret`s) using the `credentials` field on the consumer. Instead users will be able to use this (and other types of credential CRDs) to bind consumers to credentials.

In the future we'll be able to provide support for "old" way of binding - using the `credentials` field - by an implementation of an additional reconciler as in https://github.com/Kong/gateway-operator/pull/623.

czeslavo (approved):

czeslavo:
One afterthought I just had: shouldn't we name the CRD `KongCredentialBasicAuth` (with the Kong prefix) for the sake of consistency with the rest of the resources?

pmalek:
> One afterthought I just had: shouldn't we name the CRD `KongCredentialBasicAuth` (with the Kong prefix) for the sake of consistency with the rest of the resources?

I was thinking about this one and went with no prefix. If you feel that would make it more consistent (which in hindsight seems like it would) I can make the change 👍

czeslavo:
Yeah, I think we should keep it consistent. 👍

PR #87, opened 2024-09-19
feat(konnect): add KongKey

czeslavo:

Adds `KongKey` CRD along with tests.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/584.


api/configuration/v1alpha1/kongkey_types.go
```diff
@@ -0,0 +1,182 @@
...
+// KongKey is the schema for KongKey API which defines a KongKey entity.
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Programmed",description="The Resource is Programmed on Konnect",type=string,JSONPath=`.status.conditions[?(@.type=='Programmed')].status`
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec.controlPlaneRef) || has(self.spec.controlPlaneRef)", message="controlPlaneRef is required once set"
```
randmonkey, line 33: Should we add a validation rule here to disable `controlPlaneRef` to specify `namespace`?

api/configuration/v1alpha1/kongkey_types.go
```diff
@@ -0,0 +1,182 @@
...
+	// Tags is an optional set of strings associated with the Key for grouping and filtering.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
+}
+
+// PEMKeyPair defines a keypair in PEM format.
+type PEMKeyPair struct {
+	// The private key in PEM format.
+	// +kubebuilder:validation:MinLength=1
+	PrivateKey string `json:"private_key"`
```
randmonkey, line 150: Is there any requirement to redact the `PrivateKey`?

api/configuration/v1alpha1/kongkey_types.go
pmalek: @randmonkey You probably mean something like this https://github.com/Kong/kubernetes-configuration/pull/82/files#diff-134719f357e17ee676fb459e368e424fe001e8a36b50ed06128f3faf2d78d5dbR37 from #82? Yes, I believe we should be consistent with enforcing this until we have cross-namespace references designed and implemented.

api/configuration/v1alpha1/kongkey_types.go
```diff
@@ -0,0 +1,182 @@
...
+// SetConditions sets the Status Conditions.
+func (t *KongKey) SetConditions(conditions []metav1.Condition) {
+	t.Status.Conditions = conditions
+}
+
+func (t *KongKey) initKonnectStatus() {
+	t.Status.Konnect = &konnectv1alpha1.KonnectEntityStatusWithControlPlaneAndKeySetRef{}
+}
+
+// KongKeySpec defines the spec for a KongKey.
```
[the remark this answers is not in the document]
pmalek: I believe we should.

OTOH though: there is a corner case here where users could:

- specify first a CP ref and then remove that ref for a keyset ref which belongs to a different CP ref
- specify first the keyset ref and then remove that for a CP ref which is a different one that the one that keyset was bound to

To move ourselves forward I suggest that we only allow setting the CP ref for now and create an issue for implementing keyset ref binding.

I am curious to your thoughts about this.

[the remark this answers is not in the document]
czeslavo: I believe we should require the CP ref to be always set for the `Key` (as that can exist without a `KeySet`) - once set it shouldn't change as it is defined.

If the `KeySet` ref is set, we should ensure in the reconciler it refers to a `KeySet` that is bound to the same `CP` as the `Key`. If it's not, we populate a condition telling the user that the `KeySet` reference is invalid.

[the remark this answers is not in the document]
pmalek: This approach would allow users to create objects with conflicting CP refs (one direct and the other one indirect form the keyset): this would be the first object that allows this.

I'd suggest we not implement the keyset membership just now and postpone solving this until later. I'm open to other solutions though if they'd prevent admission of resources with conflicting cp refs.

api/configuration/v1alpha1/kongkey_types.go
czeslavo: Added the validation: [bda92a0](https://github.com/Kong/kubernetes-configuration/pull/87/commits/bda92a0934edbc3241cfea7e9a26a63cc36adf07)

api/configuration/v1alpha1/kongkey_types.go
czeslavo: It's one of the many fields that fall into the sensitive data bucket. I've just opened a discussion in the project's Slack channel so we can make a decision regarding them.

api/configuration/v1alpha1/kongkey_types.go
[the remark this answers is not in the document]
pmalek: @czeslavo This seems like a reasonable approach. Let's move forward with it.

Even though I do not like allowing users to create invalid objects it seems that there's no easy way to prevent this.

api/configuration/v1alpha1/kongkey_types.go
pmalek: This is the same as for other sensitive objects.

Given that we have separate types, users can enforce RBAC based on those.

api/configuration/v1alpha1/key_set_ref.go
```diff
@@ -0,0 +1,47 @@
...
+
+// KeySetRef is the schema for the KeySetRef type.
+// It is used to reference a KeySet entity.
+// +kubebuilder:validation:XValidation:rule="self.type == 'namespacedRef' ? has(self.namespacedRef) : true", message="when type is namespacedRef, namespacedRef must be set"
+// +kubebuilder:validation:XValidation:rule="self.type == 'konnectID' ? has(self.konnectID) : true", message="when type is konnectID, konnectID must be set"
+type KeySetRef struct {
+	// Type defines type of the KeySet object reference. It can be one of:
+	// - konnectID
+	// - namespacedRef
+	Type KeySetRefType `json:"type,omitempty"`
```
pmalek, line 26: Do we need the `omitempty` for anything here?

```suggestion
	Type KeySetRefType `json:"type"`
```

pmalek (approved):

api/configuration/v1alpha1/kongkey_types.go
czeslavo: Alright, let's proceed with that, at least for now. 👍

PR #95, opened 2024-09-25
add KongSNI CRD

randmonkey:
Add `KongSNI` CRD for Kong SNIs. It must be attached to a certificate.
**Which issue this PR fixes**
Required for https://github.com/Kong/gateway-operator/issues/598

**PR Readiness Checklist**:


api/configuration/v1alpha1/kongkey_types.go
pmalek, line 1: 👍

api/configuration/v1alpha1/kongsni_types.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Programmed",description="The Resource is Programmed on Konnect",type=string,JSONPath=`.status.conditions[?(@.type=='Programmed')].status`
+// +kubebuilder:validation:XValidation:rule="oldSelf.spec.certificateRef == self.spec.certificateRef", message="spec.certificateRef is immutable"
+type KongSNI struct {
```
pmalek, line 35: While there is an API that allows creating SNIs not attached to a certificate https://kong-platform-api.netlify.app/konnect/control-planes-config/v2/openapi.yaml/#tag/SNIs/operation/create-sni when using Konnect UI it's impossible to do so (see the screenshot below) and one has to specify the related certificate (and then we'd use https://kong-platform-api.netlify.app/konnect/control-planes-config/v2/openapi.yaml/#tag/SNIs/operation/create-sni-with-certificate)

image

Because of that I believe we should add the certificate ref and require it similarly how we do it for e.g. CP refs in other types.

WDYT?

randmonkey: Added a `size(self.spec.certificateRef.name) > 0` rule to limit the `certificateRef.name` to be non-empty. I am not sure if this one or `...!=''` is better.

api/configuration/v1alpha1/kongsni_types.go
```diff
@@ -0,0 +1,89 @@
...
+	Name string `json:"name"`
+
+	// Tags is an optional set of strings associated with the SNI for grouping and filtering.
+	Tags []string `json:"tags,omitempty"`
+}
+
+// KongSNISpec defines specification of a Kong SNI.
+type KongSNISpec struct {
+	// CertificateRef is the reference to the certificate to which the KongSNI is attached.
+	CertificateRef TargetRef `json:"certificateRef"`
```
czeslavo, line 59: `TargetRef` is a type defined for `KongPluginBinding`: https://github.com/Kong/kubernetes-configuration/blob/f963a53f87ac0b4cf2491f8de9d9ea9e5c464955/api/configuration/v1alpha1/kongpluginbinding_types.go#L127 We need to decide whether to make it reusable (and extract it to a separate file perhaps and rename to something more suitable the generic use case, e.g. `KongObjectRef`) or make such types separate (change the `TargetRef` to `KongPluginTargetRef` and add `KongSNICertificateRef`). I'd vote for the latter as it gives us the possibility to change their schema independently, but I'm also fine with reusing with a rename.

api/configuration/v1alpha1/kongsni_types.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Programmed",description="The Resource is Programmed on Konnect",type=string,JSONPath=`.status.conditions[?(@.type=='Programmed')].status`
+// +kubebuilder:validation:XValidation:rule="size(self.spec.certificateRef.name) > 0", message="spec.certificateRef.name is required and cannot be empty"
+// +kubebuilder:validation:XValidation:rule="oldSelf.spec.certificateRef == self.spec.certificateRef", message="spec.certificateRef is immutable"
```
czeslavo, line 35: I believe this should become immutable only since the SNI gets `Programmed`. Before that it should be fine to modify it.

czeslavo (changes requested):

api/configuration/v1alpha1/kongsni_types.go
pmalek: I'm in favor of adding generic type for this like proposed. `KongObjectRef` sounds good to me.

randmonkey: Extracted `KongObjectRef`. Would submit a separate PR to change other CRDs to use the type (and maybe also add `KongObjectRefWithGroupKind` ?)

api/configuration/v1alpha1/kongsni_types.go
randmonkey: Done.

czeslavo (approved):

Issue #31, opened 2024-08-09
Make `ControlPlaneRef` an optional pointer for non konnect-native objects


mlavacca:
## Current Behavior

Some objects that are not konnect-native (such as consumers, that are used in the context ok KIC as well), could have the `ControlPlaneRef` field migrated to an optional, `omitempty` pointer.

This way we can get rid of the default `controlPlaneRef` set to KIC. This issue is a proposal still to be discussed.

pmalek:
Reopening as not all entities have `ControlPlaneRef` set as optional yet.

pmalek:
Searching through the project it seems that we already use a pointer to `ControlPlaneRef` in every object that does have a control plane ref. 

Closing.

---

#119 adds `kic` `ControlPlaneRef` type to allow explicitly stating the cp ref type (apart from the agreed implicit default, meaning that an object is to be reconciled by KIC). cc @mlavacca

PR #98, opened 2024-09-26
feat(konnect): add KongKeySet CRD

czeslavo:

Adds `KongKeySet` CRD.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/585.

**PR Readiness Checklist**:


api/configuration/v1alpha1/kongkeyset_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+	Spec KongKeySetSpec `json:"spec"`
+
+	// +kubebuilder:default={conditions: {{type: "Programmed", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	Status KongKeySetStatus `json:"status,omitempty"`
+}
+
+// KongKeySetSpec defines the spec for a KongKeySet.
+type KongKeySetSpec struct {
+	// ControlPlaneRef is a reference to a Konnect ControlPlane this KongKeySet is associated with.
```
programmer04, line 48:
```suggestion
	// ControlPlaneRef is a reference to a Konnect ControlPlane with which KongKeySet is associated.
```

test/crdsvalidation/kongkeyset/kongkeyset_test.go
```diff
@@ -0,0 +1,66 @@
...
+					if err == nil {
+						t.Cleanup(func() {
+							assert.NoError(t, client.IgnoreNotFound(cl.Delete(ctx, entity.Name, metav1.DeleteOptions{})))
+						})
+					}
+
+					if tc.ExpectedErrorMessage == nil {
+						assert.NoError(t, err)
+
+						// if the status has to be updated, update it.
```
programmer04, line 39:
```suggestion
						// If the status has to be updated, update it.
```

test/crdsvalidation/kongkeyset/testcases/common.go
```diff
@@ -0,0 +1,34 @@
...
+	Name      string
+	TestCases []testCase
+}
+
+// TestCases is a collection of all test cases groups related to KongKeySet validation.
+var TestCases = []testCasesGroup{}
+
+func init() {
+	TestCases = append(TestCases, keySetAPISpec, cpRef)
+}
```
programmer04, line 29: Since you're specifying explicitly all tests here where the definition of `TestCases` you can add all of them as an item to a list without `init`. On the other hand `init` can be put where the test case is defined to have its registration in the same place as definition of a case. WDYT?

pmalek: I believe this can be a nice improvement but I'd stay consistent with other tests suites for now.

api/configuration/v1alpha1/kongkeyset_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// +kubebuilder:default={conditions: {{type: "Programmed", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	Status KongKeySetStatus `json:"status,omitempty"`
+}
+
+// KongKeySetSpec defines the spec for a KongKeySet.
+type KongKeySetSpec struct {
+	// ControlPlaneRef is a reference to a Konnect ControlPlane with which KongKeySet is associated.
+	// +optional
+	ControlPlaneRef *ControlPlaneRef `json:"controlPlaneRef,omitempty"`
```
pmalek, line 50: Since this is a Konnect focused type and #31 doesn't apply here and because of the fact that https://kong-platform-api.netlify.app/konnect/control-planes-config/v2/openapi.yaml/#tag/KeySets/operation/create-key-set does mark the CP ref as required I believe we should require it here as well.

pmalek (commented):
Code looks like it needs to be regenerated: https://github.com/Kong/kubernetes-configuration/actions/runs/11056910484/job/30719560512#step:6:19

api/configuration/v1alpha1/kongkeyset_types.go
czeslavo: Hmm, it makes sense. It's a net new type that no one has used so we can make the CP ref required. That will imply that if we ever decide to support it in KIC, it will only be supported with the explicit KIC CP ref.

pmalek: Hm there's still KIC's use case as you've pointed out. I'm wondering what approach we should take here: what to do with CP refs in CRDs that are not going to be used in KIC. So far I we've followed the path of making this optional (just like you have done in this PR) which allows KIC's use case with CP ref type being of KIC type by default.

Having said that I believe we can move forward with this. WDYT?

czeslavo: For sure making it optional makes our approach consistent across all CRDs. If you're ok with that, I can revert 
1c4d039 changes and leave it as was.

czeslavo: d90226d

pmalek: I believe it is for the better to keep consistent now.

api/configuration/v1alpha1/konnect_controlplaneref_types.go
```diff
@@ -19,7 +19,7 @@ type ControlPlaneRef struct {
 	// - konnectID
 	// - konnectNamespacedRef
 	// +kubebuilder:validation:Enum=konnectID;konnectNamespacedRef
-	Type string `json:"type,omitempty"`
+	Type string `json:"type"`
```
pmalek, line 22: Do we want to make this change now? I believe we want to stick with this being optional to allow implicit defaulting to KIC ( which does not have its own type's `const` defined yet).

czeslavo: This makes `Type` required when `ControlPlaneRef` is not nil, which is the right way I think therefore I wanted to fix this straight away.

pmalek: ACK, 👍

pmalek (approved):

PR #101, opened 2024-09-27
feat(konnect): allow serviceless KongRoutes

### LINKED ISSUE #99: Make `KongRoute`'s ServiceRef optional to allow serviceless routes
### Problem statement

`KongService` currently (as of https://github.com/Kong/kubernetes-configuration/releases/tag/v0.0.20) has its [ServiceRef marked as optional](https://github.com/Kong/kubernetes-configuration/blob/737bd2dcfe33bfd911e6e35bb28e4ae8cf2382c6/api/configuration/v1alpha1/kongroute_types.go#L53-L55) but there's no alternative to point to a ControlPlane to which this `KongRoute` should be bound to.

This issue tracks the effort making that change. This should include adding a CP ref which should be made an alternative to specifying the Service ref.

CEL validations should cover that.

pmalek:

Allow `KongRoute`s to be bound directly to `KonnectGatewayControlPlane`.

That is possible only when the `serviceRef` is not set. Either one of those can be set at a time.

**Which issue this PR fixes**

Fixes #99


api/configuration/v1alpha1/kongroute_types.go
```diff
@@ -36,8 +36,11 @@ import (
...
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.spec.serviceRef) || has(self.spec.serviceRef)", message="serviceRef is required once set"
-// +kubebuilder:validation:XValidation:rule="(!self.status.conditions.exists(c, c.type == 'Programmed' && c.status == 'True')) ? true : oldSelf.spec.serviceRef == self.spec.serviceRef", message="spec.serviceRef is immutable when an entity is already Programmed"
 // +kubebuilder:validation:XValidation:rule="has(self.spec.protocols) && self.spec.protocols.exists(p, p == 'http') ? (has(self.spec.hosts) || has(self.spec.methods) || has(self.spec.paths) || has(self.spec.paths) || has(self.spec.paths) || has(self.spec.headers) ) : true", message="If protocols has 'http', at least one of 'hosts', 'methods', 'paths' or 'headers' must be set"
+// +kubebuilder:validation:XValidation:rule="has(self.spec.controlPlaneRef) && !has(self.spec.serviceRef) || !has(self.spec.controlPlaneRef) && has(self.spec.serviceRef)", message="Only one of controlPlaneRef or serviceRef can be set"
```
czeslavo, line 40: What about a use case where I have a route that I first associated with a CP (no service association) and then I decided to change it to use a service? Maybe we should allow setting both CP and service refs at the same time if a route is already programmed and has the CP ref set in `oldSelf`? We'd have to cross-validate that the service is bound to the same CP in the reconciler in that case.

pmalek: That makes a lot of sense but I'd suggest that we distinguish that as a separate issue an track (and implement) it separately from the basic use case of serviceless routes. WDYT?

czeslavo: Makes sense 👍

czeslavo (approved):

PR #104, opened 2024-10-02
chore: add konnect funcs generator for konnect api

pmalek:

This extends #88 with generator for `konnect` api group as well.


scripts/konnect-funcs/supportedtypes.go
```diff
@@ -25,62 +27,94 @@ var supportedTypes = []supportedTypesT{
...
+		Types: []templateDataT{
+			{
+				Type:              "KonnectGatewayControlPlane",
+				KonnectStatusType: "*KonnectEntityStatus",
+				HasKonnectStatus:  true,
+			},
+			{
+				Type:              "KonnectAPIAuthConfiguration",
+				KonnectStatusType: "*KonnectEntityStatus",
 			},
```
czeslavo, line 118: What's the difference between an entity type with `HasKonnectStatus` set to `false` and `true` while having a `KonnectStatusType` defined? 🤔 Couldn't we rely purely on `KonnectStatusType` not being empty to determine whether we should generate the `GetKonnectStatus()` method?

pmalek: That makes sense, thanks for pointing this out. a8929c45b81df49e0882e3ba733eca50394cd7b9 should solve this.

czeslavo (approved):

PR #109, opened 2024-10-03
feat(generate): add generation of credential related funcs: `SetKonnectConsumerIDInStatus()` and `GetConsumerRefName()`

pmalek:

Add generation of credential related funcs: `SetKonnectConsumerIDInStatus()` and `GetConsumerRefName()`

### DIFF
--- a/scripts/konnect-funcs/templates.go
+++ b/scripts/konnect-funcs/templates.go
@@ -73,7 +73,19 @@ func (obj *{{ .Type }}) SetControlPlaneID(id string) {
 	}
 	obj.Status.Konnect.ControlPlaneID = id
 }
+{{- if hasSuffix "KonnectEntityStatusWithControlPlaneAndConsumerRefs" .KonnectStatusType }}
 
+func (obj *{{ .Type }}) SetKonnectConsumerIDInStatus(id string) {
+	if obj.Status.Konnect == nil {
+		obj.initKonnectStatus()
+	}
+	obj.Status.Konnect.ConsumerID = id
+}
+
+func (obj *{{ .Type }}) GetConsumerRefName() string {
+	return obj.Spec.ConsumerRef.Name
+}
+{{- end }}
 {{- end }}
 `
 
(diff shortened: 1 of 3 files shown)


scripts/konnect-funcs/templates.go
```diff
@@ -73,7 +73,19 @@ func (obj *{{ .Type }}) SetControlPlaneID(id string) {
...
+func (obj *{{ .Type }}) SetKonnectConsumerIDInStatus(id string) {
+	if obj.Status.Konnect == nil {
+		obj.initKonnectStatus()
+	}
+	obj.Status.Konnect.ConsumerID = id
+}
+
+func (obj *{{ .Type }}) GetConsumerRefName() string {
+	return obj.Spec.ConsumerRef.Name
+}
```
czeslavo, line 87: Why getting `ConsumerRefName` should depend on the `KonnectStatusType`? 

Do we need this helper anyway? Perhaps `GetKonnectConsumerID` would be more useful in terms of simplifying `nil` checks downstream. 🤔

pmalek: > Why getting `ConsumerRefName` should depend on the `KonnectStatusType`? 

Maybe this has been a bit of a shortcut on my part. WDYT about adding a separate field which would control whether a type has a consumer reference?

> Do we need this helper anyway?

The use case for this would be to simplify 
https://github.com/Kong/gateway-operator/blob/b0831b724745480c4e497d732bf978cb3f79af5f/controller/konnect/watch_credential.go#L44-L56 code and only check whether the type implements `GetConsumerRefName() string`

czeslavo: > Maybe this has been a bit of a shortcut on my part. WDYT about adding a separate field which would control whether a type has a consumer reference?

Ah! I think this shortcut should rather always be leading to truth, so I think we can leave it as is. 👍  

>The use case for this would be to simplify
https://github.com/Kong/gateway-operator/blob/b0831b724745480c4e497d732bf978cb3f79af5f/controller/konnect/watch_credential.go#L44-L56 code and only check whether the type implements GetConsumerRefName() string

Thanks for the explanation, that makes sense to me now.

czeslavo (approved):

Issue #105, opened 2024-10-03
Generate CRDs references for KGO and KIC


czeslavo:
## Problem Statement

Now that we have KIC and KGO CRDs living in this repository, we’ll need to find a way to generate sensible documentation from the API definitions we keep here (possibly separating KGO-supported and KIC-supported types so they can still be used in [KGO](https://docs.konghq.com/gateway-operator/latest/reference/custom-resources/) and [KIC](https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/custom-resources/) CRD references). 

## Proposed Solution

So far we’ve been generating the markdown files for KIC and KGO CRDs using [github.com/elastic/crd-ref-docs](http://github.com/elastic/crd-ref-docs). The idea I have is to use the same tool, but annotate the CR types with markers telling whether the CRD is supported by KIC/KGO or both (e.g. `// +kicsupported`, `// +kgosupported` on a CR type level). We can inspect the [markers](https://github.com/elastic/crd-ref-docs?tab=readme-ov-file#custom-markers) (it's a native feature of crd-ref-docs) in templates and render docs only for types that a given project supports. Templates will still live in KGO and KIC repositories.
## Acceptance Criteria

- [x] KGO's CRDs markdown reference is rendered from `kubernetes-configuration` API definitions
- [x] KIC's CRDs markdown reference is rendered from `kubernetes-configuration` API definitions

czeslavo:
KIC covered in https://github.com/Kong/kubernetes-ingress-controller/commit/0cb54115d5dd4ffc020caf18b6a3767b69866c22.

PR #110, opened 2024-10-03
chore: annotate KGO supported types

czeslavo:

Annotates all the types that should be included in the KGO docs with the new `+apireference:kgo:include` marker. Also, annotates fields that are not supported by KGO and should not be included in the docs with `+apireference:kgo:exclude`. The markers will be used to determine whether a type/field should be included in KGO's CRDs reference.

https://github.com/Kong/gateway-operator/pull/689 uses the markers.

**Which issue this PR fixes**

Part of https://github.com/Kong/kubernetes-configuration/issues/105.

### DIFF
--- a/api/configuration/v1/configsource.go
+++ b/api/configuration/v1/configsource.go
@@ -2,7 +2,6 @@ package v1
 
 // ConfigSource is a wrapper around SecretValueFromSource.
 // +kubebuilder:object:generate=true
-// +kgosupported
 type ConfigSource struct {
 	// Specifies a name and a key of a secret to refer to. The namespace is implicitly set to the one of referring object.
 	SecretValue SecretValueFromSource `json:"secretKeyRef"`
@@ -12,7 +11,6 @@ type ConfigSource struct {
 // It is an equivalent of the following patch:
 // `{"op": "add", "path": {.Path}, "value": {.ComputedValueFrom}}`.
 // +kubebuilder:object:generate=true
-// +kgosupported
 type ConfigPatch struct {
 	// Path is the JSON-Pointer value (RFC6901) that references a location within the target configuration.
 	Path string `json:"path"`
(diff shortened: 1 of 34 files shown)


api/configuration/v1/configsource.go
```diff
@@ -11,6 +12,7 @@ type ConfigSource struct {
 // It is an equivalent of the following patch:
 // `{"op": "add", "path": {.Path}, "value": {.ComputedValueFrom}}`.
 // +kubebuilder:object:generate=true
+// +kgosupported
 type ConfigPatch struct {
```
pmalek, line 16: I wonder if those types here are really considered supported by KGO 🤔.

Those are used in https://github.com/Kong/kubernetes-configuration/blob/d4d87e1acd2d5a7d333c30504c1d0aa97e6b9809/api/configuration/v1/kongplugin_types.go#L73 which is included in `kongPlugin` which is used by KGO but fields like `ConfigPatch` and `ConfigSource` are not supported in KGO.

IMHO those should not land in KGO's documentation. This can be a way of trying to hide some of the fields that may be technically available for users in KGO (because we can't add CEL rules on CRDs which are also used by KIC) but should not be set because they will have not effect.

LMKWYT

cc: @mheap

api/konnect/v1alpha1/konnect_entity_status.go
```diff
@@ -108,16 +116,18 @@ type KonnectEntityStatusWithControlPlaneAndServiceRefs struct {
 	ServiceID string `json:"serviceID,omitempty"`
 }
 
-// SetServerURL sets the server URL of the KonnectEntityStatus struct.
+// SetControlPlaneID sets the server URL of the KonnectEntityStatus struct.
 func (in *KonnectEntityStatusWithControlPlaneAndServiceRefs) SetControlPlaneID(id string) {
```
pmalek, line 120: I was trying to find a linter recently which would find those copy paste errors but I've failed 😢 

I believe we should try to either find one (which I might have missed in my search) or write one to enforce godoc comments format.

api/configuration/v1/configsource.go
czeslavo: That's a good point. I removed the marker from the types. I also had to add another marker to selectively exclude fields from the supported types (like the mentioned `ConfigPatches` and `ConfigSource` fields in `KongPlugin`).

pmalek: 👍 

Can we add a comment somewhere about how to use these markers? It can be a net new README section. WDYT?

pmalek (approved):

PR #111, opened 2024-10-03
chore: enable linter for api directory

czeslavo:

Turns the linter on for the `./api` directory, excluding only generated files. It also skips the default golanci-lint exclusions for `revive` comment rules (see: https://golangci-lint.run/usage/false-positives/#default-exclusions). 

Follow up to https://github.com/Kong/kubernetes-configuration/pull/110#discussion_r1785977244


pmalek (approved):
🏅

PR #115, opened 2024-10-07
chore: add KIC API reference markers

czeslavo:

Annotates KIC-supported `configuration` types with the new markers that are going to be used in KIC's `crd-ref-docs` templates.

**Which issue this PR fixes**

Part of https://github.com/Kong/kubernetes-configuration/issues/105.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,5 +68,8 @@ create a new tag and push it to the repository.
 |-----------------------------|------------|---------------------------------------------------------------------------------------------------------------|
 | `+apireference:kgo:exclude` | Fields     | Any field annotated with this marker will be excluded from the [KGO's generated CRDs reference][kgo-crd-ref]. |
 | `+apireference:kgo:include` | Types      | Any type annotated with this marker will be included in the [KGO's generated CRDs reference][kgo-crd-ref].    |
+| `+apireference:kic:exclude` | Fields     | Any type annotated with this marker will be excluded from the [KIC's generated CRDs reference][kic-crd-ref].  |
+| `+apireference:kic:include` | Types      | Any type annotated with this marker will be included in the [KIC's generated CRDs reference][kic-crd-ref].    |
 
 [kgo-crd-ref]: https://github.com/Kong/gateway-operator/blob/main/docs/api-reference.md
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -68,5 +68,8 @@ create a new tag and push it to the repository.
 |-----------------------------|------------|---------------------------------------------------------------------------------------------------------------|
 | `+apireference:kgo:exclude` | Fields     | Any field annotated with this marker will be excluded from the [KGO's generated CRDs reference][kgo-crd-ref]. |
 | `+apireference:kgo:include` | Types      | Any type annotated with this marker will be included in the [KGO's generated CRDs reference][kgo-crd-ref].    |
+| `+apireference:kic:exclude` | Fields     | Any type annotated with this marker will be excluded from the [KIC's generated CRDs reference][kic-crd-ref].  |
+| `+apireference:kic:include` | Types      | Any type annotated with this marker will be included in the [KIC's generated CRDs reference][kic-crd-ref].    |
```
pmalek, line 72: I wonder: what's the default action to take when KGO or KIC generate their docs? That is, is the type included or excluded when no code marker is specified?

pmalek (approved):

README.md
czeslavo: Currently, if a type is not annotated with `apireference:*:include`, it's not included (excluded by default). For fields, it's the opposite (included by default). 

Ultimately (when we can use package-level annotations), I'd like it to work like below:

1. For types, include a type if: 
- the package is annotated with `apireference:*:include-all-types` and the type is NOT annotated with `apireference:*:exclude`
OR 
- the type is annotated with `apireference:*:include`
2. For fields, include a field if the type including the field is annotated with `apireference:*:include` and the field is NOT annotated with `apireference:*:exclude`.

pmalek: Ok got it. Are we tracking the merge of https://github.com/elastic/crd-ref-docs/pull/95 somewhere so that we can introduce the package level annotations?

czeslavo: It's tracked under https://github.com/Kong/gateway-operator/issues/695.

Issue #63, opened 2024-09-04
Add CEL validation on `APIAuth` `ServerURL`


mlavacca:
## Problem Statement

The APIAuth ServerURL field has to be validated using CEL rules, according to the below requirement.

## Acceptance Criteria

- [ ] CEL validation is improved so that the ServerURL field can either be `https://<server>.<name>`, or `<server>.<name>`
- [ ] CEL validation tests are added

PR #121, opened 2024-10-09
chore: cover KonnectAPIAuthConfiguration CEL rules in tests

czeslavo:

Covers KonnectAPIAuthConfiguration CEL rules in tests. Allow `spec.serverURL` to be set to a full absolute URL with HTTPs scheme or just a hostname. 

**Which issue this PR fixes**

Fixes https://github.com/Kong/kubernetes-configuration/issues/63.


pmalek (approved):
This looks good. Thanks for adding these validations 👍

PR #122, opened 2024-10-09
feat: add list funcs

pmalek:

This PR updates the API types generator `func`s so that they additionally generate `GetItems() []T` functions for each types (in the supported lists https://github.com/Kong/kubernetes-configuration/blob/d92db6ae5d3cbba02c6eac62329c328cbbf3081c/scripts/apitypes-funcs/supportedtypes.go#L112-L114).

This will allow users of these types to e.g. bind list types and underlying item types and generically list types in watchers using something like:

```go
type ClientObject[T any] interface {
	GetName() string
	GetNamespace() string
	*T
}

func enqueueObjectForKonnectGatewayControlPlane[
	TList interface {
		client.ObjectList
		GetItems() []T
	},
	T constraints.SupportedKonnectEntityType,
	TT ClientObject[T],
](
	cl client.Client,
	index string,
) func(context.Context, client.Object) []reconcile.Request {
	return func(ctx context.Context, obj client.Object) []reconcile.Request {
		cp, ok := obj.(*konnectv1alpha1.KonnectGatewayControlPlane)
		if !ok {
			return nil
		}
		var l TList
		if err := cl.List(ctx, l,
			// TODO: change this when cross namespace refs are allowed.
			client.InNamespace(cp.GetNamespace()),
			client.MatchingFields{
				index: cp.GetNamespace() + "/" + cp.GetName(),
			},
		); err != nil {
			return nil
		}

		return objectListToReconcileRequests[T, TT](l.GetItems())
	}
}
```


test/unit/konnect_funcs_test.go
czeslavo, line 1: We could generate tests as well, WDYT?

scripts/apitypes-funcs/templates.go
```diff
@@ -130,6 +131,20 @@ func (obj *{{ .Type }}) SetConditions(conditions []metav1.Condition) {
...
 
+{{- end }}
+`
+
+	listFuncOutputFileNamme = "zz_generated_list_funcs.go"
+	listFuncTemplate        = `package {{ .PackageVersion }}
+
+// Code generated by scripts/apitypes-funcs/main.go; DO NOT EDIT.
+
+{{- range .Types }}
```
czeslavo, line 142:
```suggestion
{{ range .Types -}}
```
This way we should get the desired newline above the function comment.

scripts/apitypes-funcs/templates.go
```diff
@@ -42,7 +42,7 @@ func (obj *{{ .Type }}) SetKonnectID(id string) {
 	obj.Status.Konnect.ID = id
 }
 
-{{ end }}
+{{- end }}
```
czeslavo, line 45:
```suggestion
{{ end -}}
```
This way we should get a newline between the `GetTypeName` comment and the `SetKonnectID` function.

scripts/apitypes-funcs/templates.go
pmalek: This was necessary after all. I've changed it in: e15f8000a771b6efc7519b9e052632cb1cbd1d91

test/unit/konnect_funcs_test.go
pmalek: We could yes, but it will require a little bit more logic I think given that we need to take into account more things. 

From what I can see we could change the usage of typed clients and use `client.Client` in `test/crdsvalidation` to make it a bit less coupled to the tested type. With this we could generate tests rather easily I believe.

scripts/apitypes-funcs/templates.go
czeslavo: Hm, we still end up with the `// Code generated by scripts/apitypes-funcs/main.go; DO NOT EDIT.` comment stuck to a first `GetList` function. 🤔

pmalek: 😬 I didn't catch that. 2679cbf6ba8ca9153f2620b9e9017dc5f3ae3f78 should hopefully be it.

test/unit/konnect_funcs_test.go
czeslavo: Let's leave it for later, it's not that important. 👍

czeslavo (approved):

PR #126, opened 2024-10-11
Chore(scripts): remove unavailable "verify" make targets

randmonkey:

Remove the "verify" make targets that does not have required scripts in the repo thus unavialable.

**Which issue this PR fixes**

---

Looks like the verify items are migrated from the KIC repo but scripts are not migrated. Should we migrate the `verify-repo.sh` since it may still make sense? The `verify-versions.sh` should not be useful for this repo.

**PR Readiness Checklist**:


pmalek:
Makes sense to me to include `verify-repo.sh` 👍

randmonkey:
> Makes sense to me to include `verify-repo.sh` 👍

Added `verify.repo`.

Makefile
```diff
@@ -114,17 +114,13 @@ golangci-lint: mise yq ## Download golangci-lint locally if necessary.
...
 .PHONY: verify.diff
 verify.diff:
 	@$(PROJECT_DIR)/scripts/verify-diff.sh $(PROJECT_DIR)
 
-.PHONY: verify.versions
-verify.versions:
-	./scripts/verify-versions.sh $(TAG)
+.PHONY: verify.repo
+verify.repo:
+	./scripts/verify-repo.sh
```
pmalek, line 123: Actually, I'm wondering what's the difference between verify-repo and verify-diff. 🤔 

One is using `git diff` while the other one uses `git status`. 

After trying this out I've found that `git status --porcelain --untracked-files=no` doesn't really report an exit code other than 0 (which would trigger the `exit 1` in `verify-repo.sh`) neither with changed files nor with changed files added to index (`git add`). 

I'm wondering if we actually need it.

randmonkey: Yes, the script returns 0 when file changed. So we do not need this script and the commit  https://github.com/Kong/kubernetes-configuration/pull/126/commits/27b4b1a3a6d2355eb766fb36ca32be88bdd9493c could be removed.

pmalek (approved):

PR #130, opened 2024-10-22
feat: validate KonnectGatewayControlPlane labels

czeslavo:

Adds CEL rules for `KonnectGatewayControlPlane` `spec.labels` and covers those with tests.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/771.

### DIFF
--- a/api/konnect/v1alpha1/konnect_gateway_controlplane_types.go
+++ b/api/konnect/v1alpha1/konnect_gateway_controlplane_types.go
@@ -37,6 +37,11 @@ type KonnectGatewayControlPlane struct {
 }
 
 // KonnectGatewayControlPlaneSpec defines the desired state of KonnectGatewayControlPlane.
+// +kubebuilder:validation:XValidation:message="spec.labels must not have more than 40 entries", rule="has(self.labels) ? size(self.labels) <= 40 : true"
+// +kubebuilder:validation:XValidation:message="spec.labels keys must be of length 1-63 characters", rule="has(self.labels) ? self.labels.all(key, size(key) >= 1 && size(key) <= 63) : true"
+// +kubebuilder:validation:XValidation:message="spec.labels values must be of length 1-63 characters", rule="has(self.labels) ? self.labels.all(key, size(self.labels[key]) >= 1 && size(self.labels[key]) <= 63) : true"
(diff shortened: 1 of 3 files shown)


api/konnect/v1alpha1/konnect_gateway_controlplane_types.go
```diff
@@ -37,6 +37,11 @@ type KonnectGatewayControlPlane struct {
 }
 
 // KonnectGatewayControlPlaneSpec defines the desired state of KonnectGatewayControlPlane.
+// +kubebuilder:validation:XValidation:message="spec.labels must not have more than 40 entries", rule="has(self.labels) ? size(self.labels) <= 40 : true"
```
randmonkey, line 40: Can this item be replaced by a `MaxLength` comment on `soec.labels`?

czeslavo: Unfortunately not as `spec.labels` is defined in an SDK type. We could do that if we replaced the `sdkkonnectcomp.CreateControlPlaneRequest` with a new type defined in this repo. I'm happy to do that, as we do so for other entities already. WDYT?

randmonkey: I prefer not to do it now because it will break the implementation in KGO and we do not want to make extra changes.

api/konnect/v1alpha1/konnect_gateway_controlplane_types.go
```diff
@@ -37,6 +37,11 @@ type KonnectGatewayControlPlane struct {
...
+// +kubebuilder:validation:XValidation:message="spec.labels keys must be of length 1-63 characters", rule="has(self.labels) ? self.labels.all(key, size(key) >= 1 && size(key) <= 63) : true"
+// +kubebuilder:validation:XValidation:message="spec.labels values must be of length 1-63 characters", rule="has(self.labels) ? self.labels.all(key, size(self.labels[key]) >= 1 && size(self.labels[key]) <= 63) : true"
+// +kubebuilder:validation:XValidation:message="spec.labels keys must not start with 'kong', 'konnect', 'mesh', 'kic', 'insomnia' or '_'", rule="has(self.labels) ? self.labels.all(key, !key.startsWith('kong') && !key.startsWith('konnect') && !key.startsWith('mesh') && !key.startsWith('kic') && !key.startsWith('_') && !key.startsWith('insomnia')) : true"
```
randmonkey, line 43: actually the `_` is not required because it is limited by the following regex, but I suggest to keep it here because it is included in Konnect's error message.

randmonkey (approved):

PR #131, opened 2024-10-22
feat: validate tags in Konnect entities

czeslavo:

Adds `tags` validation to all entities supporting them. Adds `tags` to `KongConsumer` and `KongConsumerGroup` as they were missing this field.

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/771.


api/configuration/v1alpha1/kong_tags_types.go
```diff
@@ -0,0 +1,6 @@
+package v1alpha1
+
+// Tags is a type for a list of tags applied to a Kong entity.
+// +kubebuilder:validation:MaxItems=20
```
randmonkey, line 4: Kong gateway itself does not have the constraints of 20 tags and max length = 128. So the validation rules may not apply to non-Konnect environment.

czeslavo: Being more restrictive we'd still satisfy the constraints of both environments. I can see the following options:

- we implement the more restrictive rules (in this case - Konnect) and stay always safe
- we implement the less restrictive rules (in this case - Kong) and we're safe with Kong, but we fail in runtime with Konnect (and propagate errors in status)
- we do not implement any constraints on the CRD level, rely on server-side validation rules, and propagate issues in entities' status

czeslavo: I'd go with the first option as the explicit `Tags` field is added to CRDs now, so we do not break anything for users and we get a safety guarantee quite cheaply.

randmonkey: I think it is OK to apply a more restrict validation rule as it is a "new" field. If we really have such requirement to apply looser constraint (for example, using the resources to operate Kong gateway), we could update the validation rules then.

api/configuration/v1/kongconsumer_types.go
```diff
@@ -75,6 +75,9 @@ type KongConsumerSpec struct {
 	// ControlPlaneRef is a reference to a ControlPlane this Consumer is associated with.
 	// +optional
 	ControlPlaneRef *configurationv1alpha1.ControlPlaneRef `json:"controlPlaneRef,omitempty"`
+
+	// Tags is an optional set of tags applied to the consumer.
+	Tags configurationv1alpha1.Tags `json:"tags,omitempty"`
```
randmonkey, line 80: The type shared by resources in all versions may need to be moved to some common package to prevent potential cyclic dependency on packages.

czeslavo: Extracted `Tags` type to a `common` package.

randmonkey (approved):

PR #135, opened 2024-10-24
konnect: make KonnectGatewayControlPlane's cluster_type immutable

pmalek:

This PR makes `KonnectGatewayControlPlane`'s `cluster_type` immutable.

The reason for this to not be conditional on the fact whether the CP is Programmed is purely to guard users from broken state when an API request to create a CP is being made and before a successful request is being processed in the operator (and thus status condition updated), user manages to change the CP type.

### DIFF
--- a/test/crdsvalidation/konnectgatewaycontrolplane_test.go
+++ b/test/crdsvalidation/konnectgatewaycontrolplane_test.go
@@ -233,6 +233,27 @@ func TestKonnectGatewayControlPlane(t *testing.T) {
 					kcp.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name = "name-2"
 				},
 			},
+			{
+				Name: "cluster_type change is not allowed",
+				TestObject: &konnectv1alpha1.KonnectGatewayControlPlane{
+					ObjectMeta: commonObjectMeta,
+					Spec: konnectv1alpha1.KonnectGatewayControlPlaneSpec{
+						CreateControlPlaneRequest: sdkkonnectcomp.CreateControlPlaneRequest{
+							Name:        "cp-1",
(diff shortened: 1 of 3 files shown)


test/crdsvalidation/konnectgatewaycontrolplane_test.go
```diff
@@ -233,6 +233,27 @@ func TestKonnectGatewayControlPlane(t *testing.T) {
 					kcp.Spec.KonnectConfiguration.APIAuthConfigurationRef.Name = "name-2"
 				},
 			},
+			{
+				Name: "cluster_type change is not allowed",
```
czeslavo, line 237: Can we also cover the rest of the added rules (members can only be set when `CLUSTER_TYPE_CONTROL_PLANE_GROUP`, `spec.konnect.authRef` is immutable)?

pmalek: Tests for setting members are already covered in https://github.com/Kong/kubernetes-configuration/blob/b2418da9f1f569aaf882f9b5d7682ed132295eb9/test/crdsvalidation/konnectgatewaycontrolplane_test.go#L19-L132

Tests for `konnect.aurhRef` change can be found in https://github.com/Kong/kubernetes-configuration/blob/b2418da9f1f569aaf882f9b5d7682ed132295eb9/test/crdsvalidation/konnectgatewaycontrolplane_test.go#L138-L235.

Do you have something else in mind?

czeslavo: 🤦 Sorry, I misread the diff and thought these I mentioned were added.

czeslavo (approved):

PR #144, opened 2024-11-04
feat: add more plugin annotation funcs

pmalek:

This PR adds a couple of handful funcs for interacting with objects' plugins through annotations:

```
func ExtractPlugins(obj ObjectWithAnnotationsAndNamespace) []string
func ExtractPluginsNamespacedNames(obj ObjectWithAnnotationsAndNamespace) []types.NamespacedName
func ExtractPluginsWithNamespaces(obj ObjectWithAnnotationsAndNamespace) []string
```

Results from added benchmarks:

```
$go test -benchmem -bench . -run ^$ -count 1  ./pkg/metadata
goos: darwin
goarch: arm64
pkg: github.com/kong/kubernetes-configuration/pkg/metadata
cpu: Apple M1 Max
BenchmarkTestExtractPlugins/0000-10                     236752966                4.776 ns/op           0 B/op          0 allocs/op
BenchmarkTestExtractPlugins/0001-10                     21764281                56.85 ns/op           32 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0002-10                     16161171                73.78 ns/op           64 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0003-10                     12783318                91.04 ns/op           96 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0004-10                     10236519               112.7 ns/op           128 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0005-10                      8836104               134.7 ns/op           160 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0010-10                      4625467               259.5 ns/op           320 B/op          2 allocs/op
BenchmarkTestExtractPlugins/0032-10                      1633317               735.7 ns/op          1024 B/op          2 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0000-10               252518941                4.789 ns/op           0 B/op          0 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0001-10               14501283                81.82 ns/op           48 B/op          3 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0002-10                9608586               122.8 ns/op            96 B/op          4 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0003-10                6794275               174.3 ns/op           144 B/op          5 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0004-10                5497418               220.5 ns/op           192 B/op          6 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0005-10                4473290               267.3 ns/op           240 B/op          7 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0010-10                2349792               509.3 ns/op           480 B/op         12 allocs/op
BenchmarkTestExtractPluginsWithNamespaces/0032-10                 730315              1569 ns/op            1536 B/op         34 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0000-10              290015607                4.133 ns/op           0 B/op          0 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0001-10              18579889                63.00 ns/op           48 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0002-10              13824366                85.67 ns/op           96 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0003-10              10668835               116.8 ns/op           144 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0004-10               7131752               149.0 ns/op           192 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0005-10               6887876               171.7 ns/op           240 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0010-10               3150312               387.0 ns/op           480 B/op          2 allocs/op
BenchmarkTestExtractPluginsNamespacedNames/0032-10               1000000              1082 ns/op            1664 B/op          2 allocs/op
PASS
ok      github.com/kong/kubernetes-configuration/pkg/metadata   35.093s
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kong/kubernetes-configuration
 
-go 1.22.4
+go 1.23.2
 
 require (
 	github.com/Kong/sdk-konnect-go v0.1.5
--- a/pkg/metadata/plugins.go
+++ b/pkg/metadata/plugins.go
@@ -1,7 +1,6 @@
 package metadata
 
 import (
-	"iter"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -25,32 +24,6 @@ func ExtractPluginsWithNamespaces(obj ObjectWithAnnotationsAndNamespace) []strin
 // annotation set to "p1,p2" this will return []string{"p1", "p2"}
 func ExtractPlugins(obj ObjectWithAnnotationsAndNamespace) []string {
 	return extractPlugins(obj, nsOptWithoutNamespace)
-}
-
-// ExtractPluginsWithNamespacesIter extracts plugin namespaced names from the given object's
-// konghq.com/plugins annotation.
-// This function trims the whitespace from the plugin names.
-// The return value is a an iterator.
-//
-// For example, for KongConsumer in namespace default, having the "konghq.com/plugins"
-// annotation set to "p1,p2" this will return an iterator over
-// - "default/p1"
-// - "default/p2"
-func ExtractPluginsWithNamespacesIter(obj ObjectWithAnnotationsAndNamespace) iter.Seq[string] {
-	return extractPluginsIter(obj, nsOptWithNamespace)
-}
-
-// ExtractPluginsIter extracts plugin names from the given object's
-// konghq.com/plugins annotation.
-// This function trims the whitespace from the plugin names.
-// The return value is a an iterator.
-//
-// For example, for KongConsumer having the "konghq.com/plugins"
-// annotation set to "p1,p2" this will return an iterator over
-// - "p1"
-// - "p2"
-func ExtractPluginsIter(obj ObjectWithAnnotationsAndNamespace) iter.Seq[string] {
-	return extractPluginsIter(obj, nsOptWithoutNamespace)
 }
 
 // ExtractPluginsNamespacedNames extracts plugin namespaced names from the given object's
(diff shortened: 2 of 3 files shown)


czeslavo (commented):
If there are purely new and we will need to adapt downstream anyway if we'd like to use those, how about adding only the `iter` variants?

pmalek:
> If these are purely new and we will need to adapt downstream anyway if we'd like to use those, how about adding only the `iter` variants? Having too many options makes the cognitive load unnecessarily high. In this case, I believe the `iter` should always win, so why bother adding the `[]string` variants?

I wanted to provide options as each use case IMHO should decide on its own which func to use.

It might be a bit much for this particularly narrow functionality but I believe it might be worth it? I'm open to discuss this. From the benchmarks I've added it does seem that the iter flavored version is more performant but not for the cases where we have 0 or 1 plugin in the annotation.

Typically users either do not configure plugins at all or configure a relatively small number of them on an entity, let's say: less than 5 (hypothetical, nor backed by data), so using the iter approach might not make that much sense given the negligible (?) perf difference between the standard (slice based) approach.

At the very least I'd like us to have the trace of this exercise so that we know if the future what were the results.

WDYT?

pmalek:
> > Typically users either do not configure plugins at all or configure a relatively small number of them on an entity, let's say: less than 5 (hypothetical, nor backed by data), so using the iter approach might not make that much sense given the negligible (?) perf difference between the standard (slice based) approach.
> > At the very least I'd like us to have the trace of this exercise so that we know if the future what were the results.
> > WDYT?
> 
> I think that it's better to keep only one option (the non-iter one) in this lib and get back to it once we think it is necessary to improve performance in this path. Keeping two implementations for this kind of task is an overkill IMO - it's better to have one that works and not think too much about which one should be used in what situation.

Makes sense to me to limit the amount of added funcs at this stage to what's currently being used. 👍 

Removed the iter flavored funcs. PTAL.

PR #145, opened 2024-11-05
chore: add kustomization file for incubator CRDs

czeslavo:

Adds a kustomization file so it can be used externally. It's needed in KIC to replace in-tree CRDs: https://github.com/Kong/kubernetes-ingress-controller/pull/6619

### DIFF
--- a/config/crd/incubator/kustomization.yaml
+++ b/config/crd/incubator/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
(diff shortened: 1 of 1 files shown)


pmalek (approved):

pmalek (changes requested):
Do you mind adding a separate makefile target for installing those and a small test sample just like those for other types: https://github.com/Kong/kubernetes-configuration/blob/50f12a24fd51ef2fea501b09397b49f532c3555d/.github/workflows/tests.yaml#L89-L93

? This will ensure that types included in the incubator kustomization actually get "tested". 

**EDIT** seems I'm too late on this PR.

Issue #149, opened 2024-11-07
Separate channels for KIC and KGO CRDs bundles


czeslavo:
## Problem

The goal of this issue is to make it clear what is included in the CRDs bundles for KIC and KGO. Currently, for KGO we take all the CRDs (excluding the incubator package) while not all of them are supported by KGO. Same goes for KIC - not all CRDs in this repo are supported there.

## Proposed solution

- Build on top of the annotations for docs (https://github.com/Kong/kubernetes-configuration/pull/115, https://github.com/Kong/kubernetes-configuration/pull/110), possibly changing their names to not remain docs-specific (e.g. `bundle:kic:include`, `bundle:kgo:exclude` etc.).
- Generate a CRD bundle per "channel" (KIC and KGO) - an all-in-one yaml manifest with CRDs

## Acceptance criteria
- [ ] YAML bundles for KGO and KIC exist in this repository
- [ ] Bundles existence is documented in the repository's readme

PR #153, opened 2024-11-12
feat: introduce CRD channels

czeslavo:

It introduces API channels where a channel is a set of CRDs meant to be used by a specific downstream project (for now: `ingress-controller`, `ingress-contoller-incubator`, `gateway-operator`). 

Channels are defined on a CRD Go-type level using a `kong:channels` marker.

**Which issue this PR fixes**

Fixes https://github.com/Kong/kubernetes-configuration/issues/149.

**PR Readiness Checklist**:


api/configuration/v1/kongclusterplugin_types.go
```diff
@@ -40,6 +40,7 @@ import (
 
 // KongClusterPlugin is the Schema for the kongclusterplugins API.
 // +apireference:kic:include
+// +kong:channels=ingress-controller
```
randmonkey, line 43: Are we going to support it in KGO?

randmonkey (approved):

api/configuration/v1/kongclusterplugin_types.go
pmalek: That's correct. There's an issue for this here https://github.com/Kong/gateway-operator/issues/524. We decided to postpone it as it was not obvious how to model it so this will require some design work.

PR #159, opened 2024-11-15
fix: fix control plane ref validation

### LINKED ISSUE #154: Cover updates with empty ControlPlane reference
We should make sure that all types that are supported by KIC and have a ControlPlane reference can be updated with the reference being empty.

Related PR: https://github.com/Kong/kubernetes-configuration/pull/147

pmalek:

This PR:

- adds the following methods to types that have the cp ref in `configuration.konghq.com`
  - `SetControlPlaneRef(*configurationv1alpha1.ControlPlaneRef)`
  - `GetControlPlaneRef() *configurationv1alpha1.ControlPlaneRef`
- adds `NewCRDValidationTestCasesGroupCPRefChange()` which constructs `CRDValidationTestCasesGroup` in such a way that it generalizes the CRD ref tests
- fixes several CRD validation rules (which are now covered with tests mentioned above).

**Which issue this PR fixes**

Fixes #154


czeslavo (commented):
Thanks for handling this. 🙇  I like the generic approach to testing it. 

I only have one concern. Currently, KIC only supports 3 types that have a CP ref in their spec: `KongConsumer`, `KongConsumerGroup` and `KongVault`. It makes sense for them to allow a nil CP reference. But what about the rest of types? Given there's no scenario in which a CP ref being empty for a type not supported by KIC (e.g. `KongService`, `KongUpstream`, etc.) would mean anything sensible, shouldn't we require this to be set for them?

pmalek:
> I only have one concern. Currently, KIC only supports 3 types that have a CP ref in their spec: `KongConsumer`, `KongConsumerGroup` and `KongVault`. It makes sense for them to allow a nil CP reference. But what about the rest of types? Given there's no scenario in which a CP ref being empty for a type not supported by KIC (e.g. `KongService`, `KongUpstream`, etc.) would mean anything sensible, shouldn't we require this to be set for them?

Yup, that makes sense as next improvement to types that are not supported by KIC. While these would simply hang and not get reconciled by anything (with the default cp ref of `kic`) what you're proposing is going to allow faster feedback for users that could possible expect that entities of those types get translated into configuration.

I believe this could be solved by 

- adding a validation rule for spec types for those CRD that are not reconciled by KIC 
  
  ```
  // +kubebuilder:validation:XValidation:rule="!has(self.controlPlaneRef) ? true : self.controlPlaneRef.type != 'kic'", message=""
  ```

- adding a parameter to `NewCRDValidationTestCasesGroupCPRefChange` which would disable some tests for those types

- adding separate test cases for these types to ensure they cannot get cp ref type set to `kic`

WDYT?

I propose to solve this in a follow up PR as this would unnecessary make this one complex.

api/configuration/v1alpha1/zz_generated_konnect_funcs.go
czeslavo, line 1: Nit: we could change the generated file names to `zz_generated_funcs.go` as `Get/SetControlPlaneRef` is no specific to Konnect.

czeslavo (approved):

api/configuration/v1alpha1/zz_generated_konnect_funcs.go
pmalek: 👍 af2747885f9df6997b0b4e3579684d16f84b5d78

Issue #150, opened 2024-11-07
On-demand release workflow setup


czeslavo:
## Problem
Up until now, we created and pushed tags for this repository manually. Before release v1, we should make sure there's an automated, on-demand workflow that will create a proper release (tag + GH release). 

### Acceptance criteria
- [ ] An automated release workflow exists and can be triggered on demand
- [ ] A release procedure is documented in ~a release issue template~ the readme section

PR #172, opened 2024-11-26
chore: add release workflow

czeslavo:

Adds a `release.yaml` workflow. It's meant to be triggered manually to create proper GH releases.

Tested on fork: 
 - workflow run: https://github.com/czeslavo/kubernetes-configuration/actions/runs/12034412037
 - GH release: https://github.com/czeslavo/kubernetes-configuration/releases/tag/v1.0.0-rc.1

**Which issue this PR fixes**

Closes https://github.com/Kong/kubernetes-configuration/issues/150.

**PR Readiness Checklist**:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,6 +21,9 @@ jobs:
   verify-tag:
     name: Verify tag
     runs-on: ubuntu-latest
+    outputs:
+      # The prerelease part of the semver tag, if any. To be used in the release job.
+      prerelease: ${{ steps.parse-semver.outputs.prerelease }}
     steps:
       - uses: mukunku/tag-exists-action@v1.6.0
         id: check-tag
@@ -45,12 +48,6 @@ jobs:
       - tests
       - verify-tag
     steps:
-      - name: Parse semver string
-        id: parse-semver
-        uses: booxmedialtd/ws-action-parse-semver@v1.4.7
-        with:
-          input_string: ${{ github.event.inputs.tag }}
-          version_extractor_regex: 'v(.*)$'
       - uses: ncipollo/release-action@v1
         with:
           body: |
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yaml
```diff
@@ -18,6 +18,7 @@ on:
     tags:
       - '*'
   workflow_dispatch: {}
+  workflow_call: {}
```
pmalek, line 21: I didn't know that these are separate 🤔 

What does this mean? Looking at https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch seems a bit cryptic

BTW: `.github/workflows/release.yaml` above uses `workflow_dispatch` so I'm not sure I understand why that's needed.

czeslavo: `workflow_call` makes the workflow callable by other workflows (making it reusable) while `workflow_dispatch` allows triggering the workflow manually (either by a GH UI or CLI).

I added `workflow_call` to `tests.yaml` to make it possible to be called from the `release.yaml` workflow.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,73 @@
...
+    needs:
+      - tests
+      - verify-tag
+    steps:
+      - name: Parse semver string
+        id: parse-semver
+        uses: booxmedialtd/ws-action-parse-semver@v1.4.7
+        with:
+          input_string: ${{ github.event.inputs.tag }}
+          version_extractor_regex: 'v(.*)$'
```
pmalek, line 53: Since we already do this in `verify-tag` and that action has outputs we might as well use that below instead of running this action again. WDYT?

.github/workflows/tests.yaml
pmalek: 👍 Thanks for clarification. This makes sense now.

.github/workflows/release.yaml
czeslavo: Makes sense: 2e3ba05

pmalek (approved):
🎖️

PR #214, opened 2025-01-08
Revert "fix: disallow konnectID references (#198)"

czeslavo:
This reverts commit 3a54d1ac1c8e9d890f579a6408b08c094c7a2596.

It turns out that in https://github.com/Kong/kubernetes-configuration/pull/198 we didn't consider that konnectID CP reference type is supported in KGO's KonnectExtension.

### DIFF
--- a/api/configuration/v1alpha1/key_set_ref.go
+++ b/api/configuration/v1alpha1/key_set_ref.go
@@ -19,13 +19,13 @@ const (
 // KeySetRef is the schema for the KeySetRef type.
 // It is used to reference a KeySet entity.
 // +kubebuilder:validation:XValidation:rule="self.type == 'namespacedRef' ? has(self.namespacedRef) : true", message="when type is namespacedRef, namespacedRef must be set"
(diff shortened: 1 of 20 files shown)


api/configuration/v1alpha1/key_set_ref.go
```diff
@@ -19,13 +19,13 @@ const (
...
 // +kubebuilder:validation:XValidation:rule="self.type == 'namespacedRef' ? has(self.namespacedRef) : true", message="when type is namespacedRef, namespacedRef must be set"
+// +kubebuilder:validation:XValidation:rule="self.type == 'konnectID' ? has(self.konnectID) : true", message="when type is konnectID, konnectID must be set"
 // +apireference:kgo:include
 // +kong:channels=gateway-operator
 type KeySetRef struct {
 	// Type defines type of the KeySet object reference. It can be one of:
+	// - konnectID
 	// - namespacedRef
-	//
-	// +kubebuilder:validation:Enum=namespacedRef
```
pmalek, line 28: I believe we could use this enum validation but including the `konnectID`, WDYT?

czeslavo: This is already validated on the `KeySetRefType` level.

PR #216, opened 2025-01-09
chore: add contents:write permission to release workflow

czeslavo:

Should fix https://github.com/Kong/kubernetes-configuration/actions/runs/12674261856. 

Permissions reference: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release--fine-grained-access-tokens


pmalek (approved):
Thanks for taking care of this 🙇 Let's hope this fixes the workflow permissions 🤞

PR #235, opened 2025-01-17
tests: add eventually config to crd validation test suite types

pmalek:

Some tests require to check the errors with `assert.Eventually()`, for instance `ValidatingAdmissionPolicy` doesn't block invalid entries the moment the policy resource gets applied.

This PR adds this option to CRD validation test suite setup.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


czeslavo (commented):
How about using eventual assertions by default for simplicity - they will be correct in any case. If `ExpectedErrorEventuallyConfig` is not provided, we could have some sane default values. Or maybe even not make this configurable.

pmalek:
> How about using eventual assertions by default for simplicity - they will be correct in any case. If `ExpectedErrorEventuallyConfig` is not provided, we could have some sane default values. Or maybe even not make this configurable.

That makes a lot of sense 👍 How about a6d52cbe029016a15a5e0ba5f205a7571fd07d01 ?

czeslavo (approved):

PR #236, opened 2025-01-17
feat: add KongPluginBinding scope

### LINKED ISSUE #7: How to configure `KongPluginBindings` to be global
## Problem Statement

https://github.com/Kong/kubernetes-configuration/pull/1 introduced the new `KongPluginBinding` CRD to allow users to bind plugins to specific kong resources. The field `PluginRef` has been added to the `.spec` field, and such a field contains the Kind along with the plugin Name. Such a kind field is meant to be used to either refer to KongPlugin or KongClusterPlugin.

Since https://docs.konghq.com/gateway/latest/key-concepts/plugins/#precedence defines the precedence for each entity that can have plugins attached, we could think of just using the `KongPluginBinding` resource to determine the scope of a plugin. In konnect, no entity attached to a plugin means "global" and so we could do.

A separate effort is about using the `pluginBinding` to configure regular Kubernetes resources in a Gateway, but that's a separate effort described in https://github.com/Kong/kubernetes-configuration/issues/8.

## Proposed Solution

- Get rid of the `spec.pluginRef.kind` field.
- Allow kong references to be empty
- If that's the case, the plugin is global

## Acceptance Criteria

- [ ] As a user, I can easily define global plugins in Konnect

czeslavo:

Adds a `scope` field to the `KongPluginBindingSpec` with a default value of `OnlyTargets` that preserves the previous default behavior (applying a plugin to targets). Also adds `GlobalInControlPlane` enum value that allows configuring a plugin to be applied globally to a control plane.

**Which issue this PR fixes**

Closes #7. Part of https://github.com/Kong/gateway-operator/issues/440.

**PR Readiness Checklist**:

### DIFF
--- a/api/configuration/v1alpha1/kongpluginbinding_types.go
+++ b/api/configuration/v1alpha1/kongpluginbinding_types.go
@@ -22,6 +22,22 @@ import (
 	konnectv1alpha1 "github.com/kong/kubernetes-configuration/api/konnect/v1alpha1"
 )
 
+// KongPluginBindingScope defines the scope of the plugin binding.
+// Allowed values are:
+// - OnlyTargets
+// - GlobalInControlPlane
+// +kubebuilder:validation:Enum=OnlyTargets;GlobalInControlPlane
+type KongPluginBindingScope string
+
+const (
+	// KongPluginBindingScopeOnlyTargets is the scope for the plugin binding to be applied only to the targets.
+	KongPluginBindingScopeOnlyTargets KongPluginBindingScope = "OnlyTargets"
+
+	// KongPluginBindingScopeGlobalInControlPlane is the scope for the plugin binding to be applied to all entities in the
+	// control plane (a.k.a. global scope).
+	KongPluginBindingScopeGlobalInControlPlane KongPluginBindingScope = "GlobalInControlPlane"
+)
+
 // KongPluginBinding is the schema for Plugin Bindings API which defines a Kong Plugin Binding.
 //
 // +genclient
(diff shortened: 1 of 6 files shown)


api/configuration/v1alpha1/kongpluginbinding_types.go
```diff
@@ -22,6 +22,22 @@ import (
...
+// - GlobalInControlPlane
+// +kubebuilder:validation:Enum=OnlyTargets;GlobalInControlPlane
+type KongPluginBindingScope string
+
+const (
+	// KongPluginBindingScopeOnlyTargets is the scope for the plugin binding to be applied only to the targets.
+	KongPluginBindingScopeOnlyTargets KongPluginBindingScope = "OnlyTargets"
+
+	// KongPluginBindingScopeGlobalInControlPlane is the scope for the plugin binding to be applied to all entities in the
+	// control plane (a.k.a. global scope).
```
programmer04, line 37: Written like the above, we have in YAMLs 26 occurrences for `Control Plane` we have 78

programmer04 (approved):

api/configuration/v1alpha1/kongpluginbinding_types.go
programmer04: Why not? 😄  But tbh it probably will unstandardize pretty quickly (unless we enforce it). It's a nit, so it can be left without any harm

czeslavo: I'll leave it for now then.

Issue #234, opened 2025-01-17
Allow setting multiple header values in KongRoute's spec


czeslavo:
## Problem Statement

KongRoute's `spec.headers` field was initially defined as `map[string]string` while the real type expected by the Konnect API is `map[string][]string`. The goal of this issue is to allow setting multiple values of the same header. 

## Proposed Solution

As the API is already public, we should not break the schema. Because of that, adding a new `headerValues` (name to be discussed) field is proposed, deprecating the old `headers` at the same time. 

## Acceptance Criteria

- [ ] It is possible to set multiple values for a single header key in the `KongRoute`'s `spec`

PR #243, opened 2025-01-20
fix: fix KongRoute's spec.headers field type

pmalek:

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


czeslavo:
It's a breaking change to the API, shouldn't we keep the field and add a new one that will match the SDK schema? I specifically created an issue for it: https://github.com/Kong/kubernetes-configuration/issues/234

pmalek:
> It's a breaking change to the API, shouldn't we keep the field and add a new one that will match the SDK schema? I specifically created an issue for it: #234

Good point. Indeed it is a breaking change but OTOH this field, as is, has a very limited use with the fix from https://github.com/Kong/gateway-operator/pull/1063 (only 1 value allowed).

Given the small impact, users requesting the fix and the fact that introducing this breaking fix will align us with Admin API I'd propose to actually go forward and introduce the breaking change. WDYT?

czeslavo:
Considering that the API is broken in its current state, introducing a breaking change to fix it might make sense. We only have one customer who bumped into this, and we can tell them to update CRDs and KGO to address the original https://github.com/Kong/gateway-operator/issues/1043 issue.

PR #268, opened 2025-02-04
feat(konnect): add KonnectCloudGatewayNetwork CRD

### LINKED ISSUE #269: Create `KonnectNetwork` CRD
## Problem Statement

In order to create Dedicated Cloud Gateway, users have to be able to select a network.

This issue tracks the effort of adding `KonnectNetwork` CRD which will enable that.

## Acceptance Criteria

- [ ] `KonnectNetwork` CRD exists

pmalek:

Maps https://docs.konghq.com/konnect/api/cloud-gateways/latest/#/Networks/create-network API.

Ref:

- https://github.com/Kong/sdk-konnect-go/pull/63
- https://github.com/Kong/platform-api/pull/979

~This PR doesn't add any specific validations for network fields.~

After a discussion with @kevinwilson541, this API has been made immutable as networks in Konnect are considered as such.

**Which issue this PR fixes**

Fixes #269

**PR Readiness Checklist**:


api/configuration/common/controlplaneref_funcs.go
mlavacca, line 1: I think this may be problematic, as if we go this way, the structs contained in this package are not versioned anymore. Are they alpha, beta, GA? I'd lean toward keeping them inside the `v1alpha1` package.

config/crd/gateway-operator/kustomization.yaml
```diff
@@ -24,4 +24,5 @@ resources:
   - configuration.konghq.com_kongupstreams.yaml
   - configuration.konghq.com_kongvaults.yaml
   - konnect.konghq.com_konnectapiauthconfigurations.yaml
-  - konnect.konghq.com_konnectgatewaycontrolplanes.yaml
\ No newline at end of file
+  - konnect.konghq.com_konnectgatewaycontrolplanes.yaml
+  - konnect.konghq.com_konnectnetworks.yaml
```
mlavacca, line 28:
```suggestion
  - konnect.konghq.com_konnectnetworks.yaml

```

api/configuration/common/controlplaneref_funcs.go
pmalek: This was needed to prevent a cyclic import between configurationv1alpha1 and konnectv1alpha1 (where `KonnectNetwork` lands).

Do you have a different idea how to solve this? An alternative would be to add a separate apigroup package (to break the cyclic imports) and version that but that group would not be a user facing type set but rather something that would be reused in different user facing api groups.

OTOH I'm still thinking if using an unversioned `common` package would not cause problems in the future 🤔 So far I don't believe it would as long as we carefully make changes in its underlying packages.

mlavacca: We could put the `ControlPlaneRef` types inside the Konnect package. I don't think it'd create an import loop, and I think it'd make sense.

pmalek: While this might be right, it'd introduce a type that's orthogonal to Konnect inside its API group: I'm specifically thinking about the CP reference type which is not pertinent solely to Konnect but to general "configuration".

So in other words: users would have to use a konnect package to specify e.g. `KongConsumer` CP ref of type KIC. Doesn't sound right to me, even if that would be just the Go API.

pmalek: It's a reference to Kong CP, Konnect happens to be one possibility which we support. 

KIC is another (KO might change things here but let's not get ahead of ourselves here).

config/crd/gateway-operator/kustomization.yaml
pmalek: This is auto generated by `make generate`.

api/configuration/common/controlplaneref_funcs.go
mlavacca: I do have another proposal: what about putting these structs into `common/v1alpha1` and enforce the common package versioning this way?

mlavacca: let's say both CRD `foo` and `bar` use `ControlPlaneRef`. As long as both of them are alpha, `ControlPlaneRef` is in the alpha package. When `foo` or `bar` moves to beta, `ControlPlaneRef` moves to beta well. We could say that the common package version corresponds to the version of the most mature API that uses it.

Would it make sense to you?

pmalek: > let's say both CRD `foo` and `bar` use `ControlPlaneRef`. As long as both of them are alpha, `ControlPlaneRef` is in the alpha package. When `foo` or `bar` moves to beta, `ControlPlaneRef` moves to beta well. We could say that the common package version corresponds to the version of the most mature API that uses it.
> 
> Would it make sense to you?

Sounds reasonable. I'm not sure if we'll be able to enforce it down the line though.

> I do have another proposal: what about putting these structs into `common/v1alpha1` and enforce the common package versioning this way?

That would work. I'm not sure if graduating the common package (or particular types) when other types that use it graduate is feasible though.

Would you be OK with that?

mlavacca: > I'm not sure if graduating the common package (or particular types) when other types that use it graduate is feasible though.

can you expand on this? if CRD foo uses it and graduates, we'd move it the `common/v1` package and all the other CRDs references would be upgraded to use the v1 version.

pmalek: Sorry for not being clear. I meant that enforcing this will be difficult.

It will be difficult to track each usage of a type (from the  `common/<VERSION>` package) and keeping tabs on graduating it when types that do use it, graduate themselves.

mlavacca: Oh yeah, it may be, but I think it's worth trying though, rather than having no versioning at all. Would you prefer moving on with the current approach?

pmalek: I'm still on the fence but I see the value in being able to reference common/v1alpha from A/v1alpha1 and common/v1 from A/v1. 

6c9edbc69c126f633e537b51982206b78cae07c0 should do just that.

PTAL

.tools_versions.yaml
```diff
@@ -1,7 +1,7 @@
 # renovate: datasource=github-tags depName=kubernetes/code-generator
 kube-code-generator: "0.32.1"
 # renovate: datasource=github-releases depName=kubernetes-sigs/controller-tools
-controller-tools: "0.17.1"
+controller-tools: "0.17.2-0.20250205103018-dac769d27bbe"
```
pmalek, line 4: This is needed to workaround broken aliases support in controller-gen https://github.com/kubernetes-sigs/controller-tools/issues/1136 (which wasn't released yet)

mlavacca (approved):

api/konnect/v1alpha1/konnect_cloudgateway_network_types.go
```diff
@@ -0,0 +1,131 @@
...
+
+// KonnectCloudGatewayNetworkAPISpec defines the desired state of the network
+// as specified in the Konnect API contract.
+//
+// +kubebuilder:validation:XValidation:rule="self.name == oldSelf.name", message="Network name is immutable"
+// +kubebuilder:validation:XValidation:rule="self.cloud_gateway_provider_account_id == oldSelf.cloud_gateway_provider_account_id", message="Network cloud gateway provider account ID is immutable"
+// +kubebuilder:validation:XValidation:rule="self.region == oldSelf.region", message="Network region is immutable"
+// +kubebuilder:validation:XValidation:rule="self.availability_zones == oldSelf.availability_zones", message="Network availability zones are immutable"
+// +kubebuilder:validation:XValidation:rule="self.cidr_block == oldSelf.cidr_block", message="Network CIDR block is immutable"
+// +kubebuilder:validation:XValidation:rule="(!has(self.state) && !has(oldSelf.state)) || self.state == oldSelf.state", message="Network state is immutable"
```
mlavacca, line 57: Maybe we should check the programmed condition as the immutability condition? I think it would make sense to allow changes in case the resource is not programmed in Konnect yet. WDYT?

pmalek: Makes sense, let me address that.

pmalek: For some reason the validation rules added in https://github.com/Kong/kubernetes-configuration/pull/268/files#diff-4a7dbeb1a5090bccd221cc8c4541d55f93895a41ab977c1fb0aafee014f7a308R20-R27

are causing failures: https://github.com/Kong/kubernetes-configuration/actions/runs/13305937028/job/37156809821?pr=268#step:7:1876

```
KonnectCloudGatewayNetwork.konnect.konghq.com "test-cl8mh" is invalid: [<nil>: Invalid value: "object": no such key: status evaluating rule: spec.name is immutable when an entity is already Programmed,
...
```

🤔

pmalek: Ok, fixed with 09122417baf68cabe9d0dc80c575dda64c7a4a3a PTAL @mlavacca

PR #307, opened 2025-02-20
feat(konnect): add KonnectCloudGatewayDataPlaneGroupConfiguration

### LINKED ISSUE #267: Create `KonnectCloudGatewayDataPlaneGroupConfiguration` CRD
## Problem Statement

In order to support Dedicated Cloud Gateways in the operator we need to define a CRD.

This issue tracks the effort.

## Acceptance Criteria

- [ ] `DedicatedCloudGateway` CRD exists
- [ ] Analysis is performed whether additonal CRDs are needed (e.g. `NetworkConfiguration`) or single `DedicatedCloudGateway` CRD will suffice.
  - [ ] Based on the above research new CRDs are added (or not if there's no need)

pmalek:

Add `KonnectCloudGatewayDataPlaneGroupConfiguration`. 

This PR adds the API with the following constraints:

- cloud gateway network ID is only allowed to be provided via Konnect
  - this is planned to be expanded upon demand to support namespaced ref
- the API (https://docs.konghq.com/konnect/api/cloud-gateways/latest/#/Data-Plane%20Group%20Configurations/create-configuration) uses a PUT verb and enforces a single configuration against a Cloud Gateway. This means that only 1 `KonnectCloudGatewayDataPlaneGroupConfiguration` can be applied to be effective, as using more than 1 against a single Control Plane will result in configurations overriding each other
  - unless the API changes this is not planned to change

**Which issue this PR fixes**

Fixes #267

**PR Readiness Checklist**:


programmer04 (approved):

mlavacca (changes requested):
Requesting changes to have time to do a proper review

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Required
+	Value string `json:"value"`
+}
+
+// ConfigurationDataPlaneGroupAutoscale specifies the autoscale configuration for the data-plane group.
+//
+// +kubebuilder:validation:XValidation:rule="self.type == 'ConfigurationDataPlaneGroupAutoscaleStatic' ? has(self.static) : true",message="static is required when type is ConfigurationDataPlaneGroupAutoscaleStatic"
+// +kubebuilder:validation:XValidation:rule="self.type == 'ConfigurationDataPlaneGroupAutoscaleAutopilot' ? has(self.autopilot) : true",message="autopilot is required when type is ConfigurationDataPlaneGroupAutoscaleAutopilot"
+type ConfigurationDataPlaneGroupAutoscale struct {
```
mlavacca, line 153: We may want to cross-validate that when the type is autopilot, static is not set, and viceversa.

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// ConfigurationDataPlaneGroupAutoscaleAutopilot specifies the autoscale configuration for the data-plane group.
+	//
+	// +kubebuilder:validation:Optional
+	ConfigurationDataPlaneGroupAutoscaleAutopilot *ConfigurationDataPlaneGroupAutoscaleAutopilot `json:"autopilot,omitempty"`
+
+	// Type of autoscaling to use.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=ConfigurationDataPlaneGroupAutoscaleStatic;ConfigurationDataPlaneGroupAutoscaleAutopilot
```
mlavacca, line 167: Do we really need this kind of verbosity in the enum values? Could we make them a bit shorter, something like "static", "autopilot"?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+	Type sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleType `json:"type"`
+}
+
+// ConfigurationDataPlaneGroupAutoscaleAutopilot specifies the autoscale configuration for the data-plane group.
+type ConfigurationDataPlaneGroupAutoscaleAutopilot struct {
+	// Base number of requests per second that the deployment target should support.
+	//
+	// +kubebuilder:validation:Required
+	BaseRps int64 `json:"base_rps"`
```
mlavacca, line 176: Do we know of a default Konnect value that can be set here as well?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+type ConfigurationDataPlaneGroupAutoscaleStatic struct {
+	// Instance type name to indicate capacity.
+	//
+	// +kubebuilder:validation:Required
+	InstanceType sdkkonnectcomp.InstanceTypeName `json:"instance_type"`
+
+	// Number of data-planes the deployment target will contain.
+	//
+	// +kubebuilder:validation:Required
+	RequestedInstances int64 `json:"requested_instances"`
```
mlavacca, line 192: maybe we can set a default value here as well, instead of requiring users to specify it. 1 instance as default would make sense I guess?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+	// Max number of requests per second that the deployment target should support. If not set, this defaults to 10x base_rps.
+	MaxRps *int64 `json:"max_rps,omitempty"`
+}
+
+// ConfigurationDataPlaneGroupAutoscaleStatic specifies the static configuration for the data-plane group.
+type ConfigurationDataPlaneGroupAutoscaleStatic struct {
+	// Instance type name to indicate capacity.
+	//
+	// +kubebuilder:validation:Required
+	InstanceType sdkkonnectcomp.InstanceTypeName `json:"instance_type"`
```
mlavacca, line 187: can we perform some kind of validation here? Do we know the available instance types upfront?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -0,0 +1,257 @@
...
+	//
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:default={{type: "Programmed", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// KonnectCloudGatewayDataPlaneGroupConfigurationStatusGroup defines the observed state of a deployed data-plane group.
+type KonnectCloudGatewayDataPlaneGroupConfigurationStatusGroup struct {
```
mlavacca, line 217: In case this field is intended to be created and set in the same controller's iteration, I would recommend marking the fields below as required. I don't see much value in having them optional. WDYT?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: The point of making these like so was to align them with the sdk generated code in https://github.com/Kong/sdk-konnect-go/blob/08223d25420edd301cacdad9a94630a4114cd829/models/components/configurationdataplanegroupautoscale.go#L13-L16

I'm fine changing this to something more succinct like `static` and `autopilot` but I believe these will be mostly used by copy pasting and I believe aligning them with the SDK API might be worth the extra characters. WDYT?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: There's no default for this in the [SDK](https://github.com/Kong/sdk-konnect-go/blob/08223d25420edd301cacdad9a94630a4114cd829/models/components/configurationdataplanegroupautoscaleautopilot.go#L36-L37) or the OpenAPI spec itselfhttps://github.com/Kong/sdk-konnect-go/blob/d511d983a17fb1be7ddd4950cc571f4482990127/openapi.yaml#L10620-L10625 So I'd leave it as is.

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: Ditto: no default in the OAS for this field: https://github.com/Kong/sdk-konnect-go/blob/d511d983a17fb1be7ddd4950cc571f4482990127/openapi.yaml#L10602-L10605

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: I'd be inclined to add the enum here (based on https://github.com/Kong/sdk-konnect-go/blob/d511d983a17fb1be7ddd4950cc571f4482990127/openapi.yaml#L10414-L10417). What I'm worried about here is that the API might add new values and users will have to wait for us to release a new CRD version with that value added to the enum.

I'd suggest that we add a note about the current possible values (`small`, `medium` and `large`) and leave a link referring users to the API documentation e.g. https://docs.konghq.com/konnect/api/cloud-gateways/latest/#/Data-Plane%20Group%20Configurations/create-configuration. WDYT?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: I'm not sure if filling the status would all happen in the same reconciliation iteration (thinking about it, it does make sense that it would) and marking it as such could preclude e.g patching with incomplete information.


>  I don't see much value in having them optional

I do see your concern about this but at the same time: "I don't see much value in having them required". Any particular reason to make this required?

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
mlavacca: sgtm

pmalek: 👍 3aff77b6bf49a595643c5a2dff25638ce43c46a3

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: 👍 42995f14bb03334f771c11e03e546a26363710a4

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: 👍 9eaf2e981f21713caec0652eae0bf4e630ed0899

PR #315, opened 2025-02-25
feat: `KonnectExtension` `ControlPlaneRef` wrapped

mlavacca:

This PR introduces a wrap to the `ControlPlaneRef` for a specific single reason: make it future-proof. We can foresee where a `ControlPlane` could be automatically provisioned instead of being manually provided by the user. If and when that time comes, we'll be ready to put some knobs into the new struct introduced by this PR, instead of making messy changes in the broader `KonnectExtension` spec.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Adding a new version? You'll need three changes:
 - Added `konnect.konghq.com/KonnectExtension` CRD. `gateway-operator.konghq.com/KonnectExtension`
   has been deprecated.
   [#291](https://github.com/Kong/kubernetes-configuration/pull/291)
+  [#315](https://github.com/Kong/kubernetes-configuration/pull/315)
   [#317](https://github.com/Kong/kubernetes-configuration/pull/317)
 - Added `KonnectExtension` status conditions.
   [#301](https://github.com/Kong/kubernetes-configuration/pull/301)


api/konnect/v1alpha1/konnect_extension_types.go
```diff
@@ -94,6 +92,15 @@ type KonnectExtensionSpec struct {
 	DataPlaneLabels []DataPlaneLabel `json:"dataPlaneLabels,omitempty"`
 }
 
+// KonnectExtensionControlPlane is the configuration for the Konnect Control Plane.
+type KonnectExtensionControlPlane struct {
```
pmalek, line 96: Can you provide some examples of fields that would land in `KonnectExtensionControlPlane` which wouldn't fit well in the `KonnectExtensionSpec`?

mlavacca: An immediate example is the type of provisioning:
- context: running in KO (no more KIC)
- create a `GatewayConfiguration` and attach a `KonnectExtension`
- set `provisioning: automatic` in the `KonnectExtension` `ControlPlane` section
- the `Gateway`s using that GatewayConfiguration will result in a pair of `KonnectGatewayControlPlane` and a `DataPlane`. The `KonnectGatewayControlPlane` is provisioned automatically, instead of being provided by the user. All the information to do it (`KonnectAPIAuthConfiguration` reference) is already in the `KonnectExtension`

Does it make sense to you?

mlavacca: Since the type of `ControlPlane` provisioning is a `ControlPlane`-related parameter, having it in a dedicated `controlPlane`-related part of the spec, along with the `ControlPlaneRef` would make more sense than having it in the outer spec IMO.

pmalek: That makes sense, thanks 👍

pmalek (approved):
Some conflicts require resolving before merging.

mlavacca:
@pmalek PTAL, as I've dismissed your review when resolving conflicts

pmalek (approved):

PR #317, opened 2025-02-26
`KonnectExtension` dataPlane labels back to map

mlavacca:

In the previous implementation, we set the labels as an array of key-value pairs. The problem with that approach is that uniqueness of keys cannot be enforced. This PR moves back to a map with all the validation already implemented applied.

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


api/konnect/v1alpha1/konnect_extension_types.go
```diff
@@ -91,9 +91,19 @@ type KonnectExtensionSpec struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=5
-	DataPlaneLabels []DataPlaneLabel `json:"dataPlaneLabels,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key.matches('^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$'))",message="keys must match the pattern '^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$'."
+	// +kubebuilder:validation:XValidation:rule="self.all(key, !(key.startsWith('kong') || key.startsWith('konnect') || key.startsWith('insomnia') || key.startsWith('mesh') || key.startsWith('kic') || key.startsWith('_')))",message="keys must not start with 'kong', 'konnect', 'insomnia', 'mesh', 'kic', or '_'."
+	// +kubebuilder:validation:XValidation:rule="self.all(key, size(key) > 0 && size(key) < 64)",message="Too long: may not be more than 63 bytes"
+	DataPlaneLabels map[string]DataPlaneLabelValue `json:"dataPlaneLabels,omitempty"`
```
pmalek, line 97: I see that you added a separate type for the label values - `DataPlaneLabelValue` - and I 👍 that.

Could we do the same with the label names and define the validation rules on that type?

mlavacca: I've tried and it turned out that Kubebuilder does not take into account CEL rules on types used as map keys. If we define the key as a proper type and add rules on types, those won't be translated into the CRD.

On the other hand, having both keys and values as strings makes the CEL validation too expensive and the API server rejects the CRD when applied. This hybrid way overcomes both problems even if it's not the best.

pmalek: Thanks for checking. Sad to hear that :/ Let's move forward with the proposed approach then 👍

pmalek (approved):

PR #387, opened 2025-04-09
feat: introduce entity types Origin and Mirror

### LINKED ISSUE #383: Add `mirror` fields to `KonnectGatewayControlPlane`
## Problem Statement

In the context of `KonnectID` references, the `KonnectGatewayControlPlane` has to be improved to belong to different types: `origin` and `mirror`.

## Proposed Solution

Add a new `type` field to the `KonnectGatewayControlPlane` spec. This field is an enum that can have two possible values:
- `origin` (default enforced by cel)
- `mirror`

Together with the type field, a new `MirrorSpec` field has to be added: this field is a struct that contains (for the time being) only a string for the `konnectID`.

Cel validation is added to ensure that:
- in case type is origin, `MirrorSpec` is empty
- in case type is `mirror`, `MirrorSpec` is not empty and the `origin`-related fields (the ControlPlane spec) are empty

## Acceptance Criteria

- [ ] As a user, I can set a `KonnectGatewayControlPlane` to be either of `origin` type or `mirror` type.

mlavacca:

**Which issue this PR fixes**

Fixes #383 

**PR Readiness Checklist**:


api/common/v1alpha1/entitytype.go
```diff
@@ -0,0 +1,11 @@
+package v1alpha1
+
+// EntityType is the type for all the entity types.
+type EntityType string
```
randmonkey, line 4: `EntityType` looks a bit ambiguous. In the most contexts of Kong it stands for the type of an entity in Kong gateways (`services`, `rotues`, `plugins`, ...). And we are using `EntityType` in this way in many places including KIC, KGO and decK.
Since this is used for decide whether the entity is expected to be created and reconciled or only mirrored from Konnect,  can we give it a clearer and distinguishing name, like `EntitySource` or `EntityReconcileMode`?

api/common/v1alpha1/entitytype.go
```diff
@@ -0,0 +1,11 @@
...
+
+const (
+	// EntityTypeOrigin is the type for Origin entities.
+	EntityTypeOrigin EntityType = "Origin"
```
randmonkey, line 8: could we use `origin` and `mirror`? In my mind the constants usually start with lower case letters.

api/common/v1alpha1/entitytype.go
mlavacca: It makes sense. I've used `EntitySource` 👍🏻

api/common/v1alpha1/entitytype.go
mlavacca: The upstream api convention states that the allowed values for the enum fields should be `CamelCase`: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

randmonkey (approved):

Issue #246, opened 2025-01-21
Use KAL linter for Kubernetes API definitions


czeslavo:
## Problem Statement

Defining Kubernetes API via Go types has its pitfalls. Some of them can be caught using a new KAL linter: https://github.com/JoelSpeed/kal. 

Inspired by https://github.com/Kong/kubernetes-configuration/pull/238#discussion_r1923435692.

## Proposed Solution

Use KAL as a standalone binary (as of now, it's not part of the golangcilint suite, so that should be the easiest way for now - we can migrate once it's included).

## Acceptance Criteria

- [ ] All Kubernetes API types defined in this repository are linted using https://github.com/JoelSpeed/kal

PR #441, opened 2025-05-22
chore: add v2alpha1 group and add ControlPlane type

pmalek:

**Which issue this PR fixes**

Part of https://github.com/Kong/gateway-operator/issues/1358

**PR Readiness Checklist**:


api/gateway-operator/v2alpha1/doc.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2025 Kong, Inc.
```
programmer04, line 2: Here date is ok

api/gateway-operator/v2alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2022 Kong Inc.
```
programmer04, line 2: But here not

pmalek:
```suggestion
Copyright 2025 Kong Inc.
```

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	IngressClass *string `json:"ingressClass,omitempty"`
+}
+
+// ControlPlaneOptions indicates the specific information needed to
+// deploy and connect a ControlPlane to a DataPlane object.
+//
+// +apireference:kgo:include
+// +kubebuilder:validation:XValidation:message="Extension not allowed for ControlPlane",rule="has(self.extensions) ? self.extensions.all(e, (e.group == 'konnect.konghq.com' && e.kind == 'KonnectExtension') || (e.group == 'gateway-operator.konghq.com' && e.kind == 'DataPlaneMetricsExtension')) : true"
+type ControlPlaneOptions struct {
+	// DataPlanes designates the target data plane to configure.
```
tao12345666333, line 94:
```suggestion
	// DataPlane designates the target data plane to configure.
```

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	DataPlane ControlPlaneDataPlaneTarget `json:"dataplane,omitempty"`
+
+	// Extensions provide additional or replacement features for the ControlPlane
+	// resources to influence or enhance functionality.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=0
+	// +kubebuilder:validation:MaxItems=2
+	Extensions []commonv1alpha1.ExtensionRef `json:"extensions,omitempty"`
+
```
tao12345666333, line 110: Is this because we only have 2 extensions?

tao12345666333 (approved):

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	URL string `json:"url,omitempty"`
+
+	// Name is the name of the DataPlane to configure.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
+}
+
+// ControlPlaneAdminAPI defines the configuration for the Kong Admin API that
+// a ControlPlane when configuring the DataPlane.
```
mlavacca, line 180: I guess we missed some words here?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+
+	// Enabled indicates whether the controller is enabled or not.
+	//
+	// +kubebuilder:validation:Required
+	Enabled *bool `json:"enabled,omitempty"`
+}
```
mlavacca, line 204: I don't think this is great UX: let's say a user is running a CP with X and Y controllers explicitly enabled. How this struct works does not define the current state of the world: as a user, I should go to the KGO docs, see which controllers are enabled by default, and cross that data with the one here defined. OTOH, I understand the need to have such an implicit default. We could add some information in the status to list all the controllers actually enabled for the specific instance, to give users a good feedback on what's going on. WDYT?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// Name is the name of the feature gate.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+
+	// Enabled indicates whether the feature gate is enabled or not.
+	//
+	// +kubebuilder:validation:Required
+	Enabled *bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
mlavacca: We should consider if we want to follow the upstream approach:
- alpha: FG disabled by default
- beta: FG enabled by default, can be turned off
- GA: no FG, the feature is on and cannot be turned off

If we want to have the same level of expressiveness, I think we should keep the possibility to disable FGs, otherwise, we end up having just alpha and GA features.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// ControlPlaneFeatureGate defines a feature gate state for the ControlPlane.
+// It overrides the default behavior as defined in the deployed operator version.
+//
+// +apireference:kgo:include
+type ControlPlaneFeatureGate struct {
```
mlavacca, line 210: Very similarly to the enabled controllers field, I think it could be super beneficial to have proper status feedback about what FGs are enabled.

mlavacca (changes requested):
Thanks for the effort on this, @pmalek!
I've left a few comments, curious to hear your opinion :)

requesting changes to avoid accidental merge.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	//
+	// +kubebuilder:validation:Enum=url;name
+	// +kubebuilder:validation:Required
+	Type string `json:"type,omitempty"`
+
+	// URL is the URL of the DataPlane target. This is used for configuring
+	// externally managed DataPlanes like those installed independently with Helm.
+	//
+	// +optional
+	URL string `json:"url,omitempty"`
```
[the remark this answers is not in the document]
pmalek: Yes, good catch. 2946df242484dc1b061844f0f1b8df6a98386d55

api/gateway-operator/v2alpha1/controlplane_types.go
pmalek: Yup, f287f8b87209a0db2fecfcbb64ab9a5530ee3036 should fix that.

api/gateway-operator/v2alpha1/controlplane_types.go
pmalek: I agree. This is a good idea to have this in status. Let me work on that

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	SchemeBuilder.Register(&ControlPlane{}, &ControlPlaneList{})
+}
+
+// ControlPlane is the Schema for the controlplanes API
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=kocp,categories=kong;all
```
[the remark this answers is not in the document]
pmalek: I agree on this and I've created #445 to track this. Feel free to amend it so that it expresses your concerns accurately.

My concerns about this is that it might not get the priority on the roadmap, or not get clear expectations defined (e.g. which categories should be assigned to which resources).

Having said that I do believe there is something to be done in this subject.

docs/gateway-operator-api-reference.md
```diff
@@ -1299,6 +1300,8 @@ WatchNamespaces defines the namespaces to watch for resources
 
 _Appears in:_
 - [ControlPlaneOptions](#controlplaneoptions)
+- [ControlPlaneOptions](#controlplaneoptions)
```
sandromodarelli, line 1303: we should use an anchor for #controlplaneoptions that belongs to the proper CRD API version

docs/all-api-reference.md
```diff
@@ -3368,6 +3371,150 @@ _Appears in:_
...
+| `controllers` _[ControlPlaneController](#controlplanecontroller) array_ | Controllers defines the controllers that are enabled for this ControlPlane. |
+| `adminAPI` _[ControlPlaneAdminAPI](#controlplaneadminapi)_ | AdminAPI defines the configuration for the Kong Admin API. |
+| `gatewayClass` _[ObjectName](#objectname)_ | GatewayClass indicates the Gateway resources which this ControlPlane should be responsible for configuring routes for (e.g. HTTPRoute, TCPRoute, UDPRoute, TLSRoute, etc.).<br /><br /> Required for the ControlPlane to have any effect: at least one Gateway must be present for configuration to be pushed to the data-plane and only Gateway resources can be used to identify data-plane entities. |
+| `ingressClass` _string_ | IngressClass enables support for the older Ingress resource and indicates which Ingress resources this ControlPlane should be responsible for.<br /><br /> Routing configured this way will be applied to the Gateway resources indicated by GatewayClass.<br /><br /> If omitted, Ingress resources will not be supported by the ControlPlane. |
+
+
+_Appears in:_
+- [ControlPlane](#controlplane)
```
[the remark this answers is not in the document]
pmalek: Well spotted. This seems to be an issue with `crd-ref-docs` which we're using and there's already an issue upstream for this https://github.com/elastic/crd-ref-docs/issues/60 and a PR that fixes it https://github.com/elastic/crd-ref-docs/pull/118 but it doesn't work quite yet.

I've created an issue in this project to track this (for KO 2.0): #446.

My suggestion is to handle the API changes here and tackle the documentation issues separately, WDYT @sandromodarelli ?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,265 @@
...
+	//
+	// +kubebuilder:validation:Enum=url;name
+	// +kubebuilder:validation:Required
+	Type ControlPlaneDataPlaneTargetType `json:"type,omitempty"`
+
+	// URL is the URL of the DataPlane target. This is used for configuring
+	// externally managed DataPlanes like those installed independently with Helm.
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https?://[a-zA-Z0-9.-]+(:[0-9]+)?(/.*)?$`
```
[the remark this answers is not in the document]
pmalek: Good call. Indeed we can use `isURL` here. 93b36efe3641d84e5a8329c46a032dd98392eb7b does that. PTAL

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,265 @@
...
+	// managed by the operator.
+	//
+	// +required
+	DataPlane ControlPlaneDataPlaneTarget `json:"dataplane,omitempty"`
+
+	// Extensions provide additional or replacement features for the ControlPlane
+	// resources to influence or enhance functionality.
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=0
```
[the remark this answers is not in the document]
pmalek: It's not necessary. 93b36efe3641d84e5a8329c46a032dd98392eb7b removes that.

> Also is there any unique constraint to limit that we cannot have 2 extensions with the same group and kind?

This is a good idea to pursue but I'm not sure if this is feasible. Even if wanted to try to run a doubly nested loop (with O(n^2) complexity) it might be rejected by API server due to cost restrictions. Because of that I'd suggest to not try to go this route.

Do you have a different idea how to tackle this @randmonkey ?

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
pmalek: The field is a `*bool` with required marker to only 2 states are allowed, enabled and disabled (any other combination will be rejected by the API server). If you want to use the default behavior you don't define anything for that field.

docs/gateway-operator-api-reference.md
```diff
@@ -1299,6 +1300,8 @@ WatchNamespaces defines the namespaces to watch for resources
...
+- [ControlPlaneSpec](#controlplanespec)
```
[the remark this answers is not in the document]
pmalek: See my comment in https://github.com/Kong/kubernetes-configuration/pull/441#discussion_r2126097550

docs/gateway-operator-api-reference.md
pmalek: Please see my comment in https://github.com/Kong/kubernetes-configuration/pull/441#discussion_r2126097550

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
mlavacca: Thanks for opening that issue! Let me add an item for next team sync meeting to make sure it gets discussed and prioritized 👍🏻

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// ControlPlane is the Schema for the controlplanes API
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=kocp,categories=kong;all
+// +kubebuilder:printcolumn:name="Ready",description="The Resource is ready",type=string,JSONPath=`.status.conditions[?(@.type=='Ready')].status`
```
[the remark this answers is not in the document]
mlavacca: That makes sense. During the implementation, I guess we'll have to surface this information from the controller routine to the CP controller.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// GatewayClass indicates the Gateway resources which this ControlPlane
+	// should be responsible for configuring routes for (e.g. HTTPRoute,
+	// TCPRoute, UDPRoute, TLSRoute, etc.).
+	//
+	// Required for the ControlPlane to have any effect: at least one Gateway
+	// must be present for configuration to be pushed to the data-plane and
+	// only Gateway resources can be used to identify data-plane entities.
+	//
+	// +optional
+	GatewayClass *gatewayv1.ObjectName `json:"gatewayClass,omitempty"`
```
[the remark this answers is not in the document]
mlavacca: 100% agree. I think we should set the Gateway name here to reflect this 1:1 relationship.  In this case, I think we should remove the `--gateway-to-reconcile` flag to avoid users shooting themselves in the foot by setting both the fields.

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
randmonkey: So why should it be a pointer but noted `required` and `omitempty`? I think it is not meaningful to define the field to a pointer and add an `omitempty` annotation as it is required.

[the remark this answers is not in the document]
mlavacca: that makes sense to me. Let me summarize what you said here to make sure we are on the same page:

| | alpha     | beta      | GA |
|--|------------- | ------------- | ---|
| true | enabled | enabled | enabled |
| false | disabled | disabled | enabled |
| nil | disabled | enabled | enabled |

On GA features, there is no gonna be any FG, so the feature is enabled regardless of the FGs enabled.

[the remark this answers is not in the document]
pmalek: Ah very well spotted. When we introduce kube-api-linter at some point it should catch these problems (we already have https://github.com/Kong/kubernetes-configuration/issues/246 for this).

f357eb47acadfd79061e8d55de4600e266d7d62e removes the omitempty. PTAL

docs/all-api-reference.md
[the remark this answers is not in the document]
sandromodarelli: agree 👍

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,264 @@
...
+	// +kubebuilder:validation:MaxItems=32
+	FeatureGates []ControlPlaneFeatureGate `json:"featureGates,omitempty"`
+
+	// Controllers is a list of enabled and disabled controllers for this ControlPlane.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=32
+	Controllers []ControlPlaneController `json:"controllers,omitempty"`
```
mlavacca, line 248: Since the status reflects the current status of the CP, what do you think about setting here just the enabled controllers? I'm just thinking of a list of plain strings with the controller names.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,264 @@
...
+	// +kubebuilder:default={{type: "Scheduled", status: "Unknown", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// FeatureGates is a list of effective feature gates for this ControlPlane.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=32
+	FeatureGates []ControlPlaneFeatureGate `json:"featureGates,omitempty"`
```
mlavacca, line 240: ditto about setting only the enabled ones.

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
randmonkey: Looks like `x-kubernetes-list-map-keys` does not support combination of multiple fields of elements in the list so it is OK to not go this way.

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
pmalek: Well, `--gateway-to-reconcile` is the CLI flag which only makes sense in KIC's context. What I proposed here is to set its backing configuration knob through the `gateway` field so I guess we're aligned on this here, right?

But now that we're discussing it I see a small problem: with the `dataplane` field users will be able to set `dataplane.name` to a conflicting value with the `gateway` field ( the selected `DataPlane` can be selected as one that is not managed by the selected `Gateway` ). So I think that rather than adding a `gateway` field, we add a third option to `dataplane.type`, something like: `managedByGateway` and users would have to provide the `Gateway` name as value there. Does that make sense?

api/gateway-operator/v2alpha1/controlplane_types.go
pmalek: On one hand that makes sense but OTOH having also the disabled controllers (or feature gates) makes it very easy for users to see which items they can enable in the spec. Without this information they have to consult the documentation.

WDYT?

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
mlavacca: > Well, `--gateway-to-reconcile` is the CLI flag which only makes sense in KIC's context. What I proposed here is to set its backing configuration knob through the `gateway` field so I guess we're aligned on this here, right?

yep, we are on the same page 👍🏻 

> But now that we're discussing it I see a small problem: with the `dataplane` field users will be able to set `dataplane.name` to a conflicting value with the `gateway` field ( the selected `DataPlane` can be selected as one that is not managed by the selected `Gateway` ). So I think that rather than adding a `gateway` field, we add a third option to `dataplane.type`, something like: `managedByGateway` and users would have to provide the `Gateway` name as value there. Does that make sense?

Yes, this is definitely a problem we have to think about.
I think that we need to provide both the DP and the Gateway name, as they are required to perform different operations. We could approach the problem as follows:
- make dataplane and gateway mutually exclusive
- in case the gateway is set, we infer the dataplane name by label (the DP has the `managed-by` label)
- we set the dataplane name on the CP status, regardless it's explicitly set by the user or inferred by the CP operator
- the DP in the CP status is the source of truth for the subsequent operations.

WDYT?

api/gateway-operator/v2alpha1/controlplane_types.go
mlavacca: Makes sense. So, your opinion is to list here ALL the available controllers along with their enablement?

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
mlavacca: yep, looks good to me 👍🏻

api/gateway-operator/v2alpha1/controlplane_types.go
pmalek: All, as defined in particular operator version. So whatever the operator is aware of. This will allow the user to know just by looking at the status what is available for enabling/disabling.

I'll then go ahead and make the change.

docs/all-api-reference.md
```diff
@@ -3364,6 +3367,164 @@ _Appears in:_
...
+is responsible for configuring.
+
+
+
+| Field | Description |
+| --- | --- |
+| `type` _[ControlPlaneDataPlaneTargetType](#controlplanedataplanetargettype)_ | Type indicates the type of the DataPlane target. |
+| `url` _string_ | URL is the URL of the DataPlane target. This is used for configuring externally managed DataPlanes like those installed independently with Helm. |
+| `name` _string_ | Name is the name of the DataPlane to configure. |
+
```
sandromodarelli, line 3444: what do you think about adding as spec here the `--kong-admin-*` KIC parameters?

pmalek: This will automatically populate when we define the external dataplane spec field: https://github.com/Kong/kubernetes-configuration/pull/441/files#diff-615b50895f2f01d1d99e673686c9c067a9329d301f9a14af438b5a890574a15bR171-R182.

I've intentionally left this incomplete to specify all the other options in a smaller PR, more focused on tackling https://github.com/Kong/gateway-operator/issues/1366.

Does that sound good to you @sandromodarelli?

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
pmalek: We've ended up discussing the API with @mlavacca and we came up with the following 3 use cases on how the ControlPlane can be used:

```
# Use case: created by user together with DataPlane.
#
# Configuration API: just Ingress.
# Require the ingress class field.
apiVersion: gateway-operator.konghq.com/v2alpha1
kind: ControlPlane
metadata:
  name: controlplane-v2-example-1
spec:
  ingressClass: kong # optional, not defaulted!
  dataplane:
    type: ref
    ref:
      name: dataplane-name
---
# Use case: created by the operator as part of the Gateway. Ownership metadata
# used to get the owner Gateway resource.
#
# Configuration API:
# - Gateway API only through the Gateway ref. Ingress resources disabled.
# - Gateway API and Ingress if IngressClass param is provided.
apiVersion: gateway-operator.konghq.com/v2alpha1
kind: ControlPlane
metadata:
  name: controlplane-v2-example-2
spec:
  # ingressClass: kong # optional, not defaulted!
  dataplane:
    type: managedByOwner
---
# TODO: discuss with product.
# Configuration API ? Only Ingress?
# Usecase ?
apiVersion: gateway-operator.konghq.com/v2alpha1
kind: ControlPlane
metadata:
  name: controlplane-v2-example-3
spec:
  dataplane:
    type: external
    external:
      url: https://10.0.8.3:8444/admin
      # auth: secret | password
      # secret: kong-admin-password-secret-name
      # password: kong-admin-password
```

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,301 @@
...
+	// ControlPlaneDataPlaneTargetRefType indicates that the DataPlane target is a ref
+	// of a DataPlane resource managed by the operator.
+	// This is used for configuring DataPlanes that are managed by the operator.
+	ControlPlaneDataPlaneTargetRefType ControlPlaneDataPlaneTargetType = "ref"
+
+	// ControlPlaneDataPlaneTargetManagedByType indicates that the DataPlane target
+	// is managed by the owner of the ControlPlane.
+	// This is the case when using a Gateway resource to manage the DataPlane
+	// and the ControlPlane is responsible for configuring it.
+	ControlPlaneDataPlaneTargetManagedByType ControlPlaneDataPlaneTargetType = "managedByOwner"
```
pmalek, line 168: I've tried constraining the created `ControlPlane` with this target type to ensure that it has an owner set but CEL doesn't allow it for some reason:

```
The CustomResourceDefinition "controlplanes.gateway-operator.konghq.com" is invalid: spec.versions[1].schema.openAPIV3Schema.x-kubernetes-validations[0].rule: Invalid value: apiextensions.ValidationRule{Rule:"self.spec.dataplane.type != 'managedByOwner' || self.ownerReferences[0].kind == 'Gateway'", Message:"X", MessageExpression:"", Reason:(*apiextensions.FieldValueErrorReason)(nil), FieldPath:"", OptionalOldSelf:(*bool)(nil)}: compilation failed: ERROR: <input>:1:53: undefined field 'ownerReferences'
 | self.spec.dataplane.type != 'managedByOwner' || self.ownerReferences[0].kind == 'Gateway'
 | ....................................................^
```

slack thread asking about the reason and the comprehensive list of allowed `metadata` fields to use in CEL: https://kubernetes.slack.com/archives/C0EG7JC6T/p1749212565802129

docs/all-api-reference.md
sandromodarelli: that's perfect 🚀

mlavacca (commented):
Mostly lgtm, just one nit about validation

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -0,0 +1,301 @@
...
+	AdminAPI *ControlPlaneAdminAPI `json:"adminAPI,omitempty"`
+}
+
+// ControlPlaneDataPlaneTarget defines the target for the DataPlane that the ControlPlane
+// is responsible for configuring.
+//
+// +kubebuilder:validation:XValidation:message="External has to be provided when type is set to external",rule="self.type != 'external' || has(self.external)"
+// +kubebuilder:validation:XValidation:message="External cannot be provided when type is set to managedByOwner",rule="self.type != 'managedByOwner' || !has(self.external)"
+// +kubebuilder:validation:XValidation:message="Ref has to be provided when type is set to ref",rule="self.type != 'ref' || has(self.ref)"
+// +kubebuilder:validation:XValidation:message="Ref cannot be provided when type is set to managedByOwner",rule="self.type != 'managedByOwner' || !has(self.ref)"
```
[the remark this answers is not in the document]
pmalek: Thanks for catching this. Addressed in 804f3bf9c28973307676769f796091e260ae91e7. 

PTAL.

api/gateway-operator/v2alpha1/controlplane_types.go
pmalek: This is explained in https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/

> The apiVersion, kind, metadata.name and metadata.generateName are always accessible from the root of the object and from any x-kubernetes-embedded-resource annotated objects. No other metadata properties are accessible.

So we can't do this validation (at this moment) via CEL.

PR #464, opened 2025-06-17
deprecate: KongIngress, TCPIngress, UDPIngress

tao12345666333:

**Which issue this PR fixes**

part of https://github.com/Kong/kubernetes-ingress-controller/issues/7496

**PR Readiness Checklist**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@ Adding a new version? You'll need three changes:
   [#450](https://github.com/Kong/kubernetes-configuration/pull/450)
 - `DataPlaneMetricsExtension` is not marked as EE only anymore.
   [#456](https://github.com/Kong/kubernetes-configuration/pull/456)
+- Marked `KongIngress`, `TCPIngress`, and `UDPIngress` as deprecated.
+  You can use Gateway API resources instead.
+  <!--- TODO(Jintao): Add docs link to Deprecated description once we have a
+  documentation page for the migration.
+  https://github.com/Kong/kubernetes-ingress-controller/issues/7496 --->
+  [#464](https://github.com/Kong/kubernetes-configuration/pull/464)
 
 ### Breaking Changes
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ Adding a new version? You'll need three changes:
   [#450](https://github.com/Kong/kubernetes-configuration/pull/450)
 - `DataPlaneMetricsExtension` is not marked as EE only anymore.
   [#456](https://github.com/Kong/kubernetes-configuration/pull/456)
+- Marked `KongIngress`, `TCPIngress`, and `UDPIngress` as deprecated.
+  You can use Gateway API resources instead.
```
pmalek, line 45: I think we should be more specific in giving the Gateway API alternatives here.

- If I'm a user of `TCPIngress` what options do I have now with Gateway API?
- If I'm a user of `KongIngress` what options do I have now with Gateway API?
- If I'm a user of `UDPIngress` what options do I have now with Gateway API?

Without this I think that users might feel lost in how they should proceed. (this concerns both the changelog entry and the godocs on types)

tao12345666333: I think this detailed migration path should be included in the documentation, what do you think?

This item is also included in the AC of https://github.com/Kong/kubernetes-ingress-controller/issues/7496.

> Available documentation about migration from deprecated CRDs to related GWAPI resources

pmalek: Fair enough. In that case I'd make sure to add a TODO comment in the changelog and Go types comments with placeholder docs links so that we don't forget about those.

Additionally, the `fixes` verb in the PR description will have to change to not close 7496 when we merge this here (since more work will need to be done to add these migration doc links).

tao12345666333: I changed it to `part of`

tao12345666333: @pmalek https://github.com/Kong/kubernetes-configuration/pull/464/commits/f349d8e55dab81963e6fdb1ce11d6d4af6fcb34d added TODOs

pmalek (approved):

PR #513, opened 2025-07-16
feat(controlplane): Add spec.dataplaneSync in ControlPlane CRD

randmonkey:
Add `spec.dataplaneSync` in `ControlPlane` CRD to configure the options to sync configuration with managed dataplanes.
Includes `interval` to configure interval between two successive syncs and `timeout` to set timeout of a single sync.
**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/issues/1805

- Should we limit the lower bound of `interval` to prevent too frequent syncs?
**PR Readiness Checklist**:


api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -147,6 +147,14 @@ type ControlPlaneDataPlaneSync struct {
 	// +optional
 	// +kubebuilder:default=false
 	EnableReverse *bool `json:"enableReverse,omitempty"`
+	// Interval is the interval between two rounds of syncing Kong configuration with dataplanes.
+	//
+	// +optional
+	Interval *metav1.Duration `json:"interval,omitempty"`
```
randmonkey, line 153: REVIEW: set a minimal value and a default value for the interval?

programmer04: I think that the same value as in KIC should be used

PR #534, opened 2025-07-29
feat(controlplane): Add `spec.objectFilters` to limit watched secrets and configMaps 

randmonkey:

Add `spec.objectFilters` in `ControlPlane` to limit watched secrets and configmaps by `secrets.allowedLabels` and `configMaps,allowedLabels`.

**Which issue this PR fixes**

CRD update part of https://github.com/Kong/kong-operator/issues/1863

**PR Readiness Checklist**:


api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -352,6 +357,28 @@ type ControlPlaneConfigDump struct {
...
+	// +optional
+	Secrets *ControlPlaneFilterForSingleObject `json:"secrets,omitempty"`
+	// ConfigMaps defines the filters for watched config maps.
+	//
+	// +optional
+	ConfigMaps *ControlPlaneFilterForSingleObject `json:"configMaps,omitempty"`
+}
+
+// ControlPlaneFilterForSingleObject defines the filters for a certain type of object.
+type ControlPlaneFilterForSingleObject struct {
```
pmalek, line 373: I know naming is hard but I believe we could a bit better hear to get a clearer name.

WDYT about:

```suggestion
// ControlPlaneFilterForObjectType defines the filters for a certain type of object.
type ControlPlaneFilterForObjectType struct {
```

?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -352,6 +357,28 @@ type ControlPlaneConfigDump struct {
...
+	// AllowedLabel specifies a label that an object must have the label with value "true" to get watched by the controllers.
+	// For example, if the secrets.allowedLabel is set to "some-label", only secrets with label "some-label=true" are reconciled.
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	AllowedLabel *string `json:"allowedLabel,omitempty"` // REVIEW: add other validation of label keys, like the prefix/name format, max length, used chars?
```
pmalek, line 379: IMHO allowed doesn't have the accurate meaning for this use case, WDYT about using:

```suggestion
	// MustHaveLabel specifies a label that an object must have set to "true" to be watched by the controllers.
	// For example, if the secrets.mustHaveLabel is set to "some-label", only secrets with label "some-label=true" are reconciled.
	//
	// +optional
	// +kubebuilder:validation:MinLength=1
	MustHaveLabel *string `json:"mustHaveLabel,omitempty"`
```

>  // REVIEW: add other validation of label keys, like the prefix/name format, max length, used chars?

I think we can solve that when there's user/product demand for this.

api/gateway-operator/v2alpha1/controlplane_types.go
randmonkey: Done.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -352,6 +357,28 @@ type ControlPlaneConfigDump struct {
...
+}
+
+// ControlPlaneFilterForObjectType defines the filters for a certain type of object.
+type ControlPlaneFilterForObjectType struct {
+	// MustHaveLabel specifies a label that an object must have the label with value "true" to get watched by the controllers.
+	// For example, if the secrets.mustHaveLabel is set to "some-label", only secrets with label "some-label=true" are reconciled.
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	MustHaveLabel *string `json:"mustHaveLabel,omitempty"` // REVIEW: add other validation of label keys, like the prefix/name format, max length, used chars?
```
pmalek, line 379: > // REVIEW: add other validation of label keys, like the prefix/name format, max length, used chars?

I think we'll want to add that when we get product / user demand for this.

Perhaps a good option for making the API extensible would be to make `ControlPlaneFilterForObjectType` a union type, i.e. add a `filterType` (or better named) field which would indicate which filtering option is chosen. Initially only the mustHabeLabel would be supported but it would allow us to extend this API in the future.

```
configMaps:
  filterType: mustHaveLabel
  mustHaveLabel:
    name: my-label
```

WDYT?

PR #535, opened 2025-07-29
feat(controlplane): add konnect related configuration fields

tao12345666333:

**Which issue this PR fixes**

part of https://github.com/Kong/kong-operator/issues/1804

**PR Readiness Checklist**:


api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -458,3 +463,82 @@ func (c *ControlPlane) SetConditions(conditions []metav1.Condition) {
...
 }
+
+// ControlPlaneKonnectOptions defines the Konnect-related configuration options for the ControlPlane.
+//
+// +apireference:kgo:include
+type ControlPlaneKonnectOptions struct {
+	// ConsumersSync defines the configuration for synchronizing consumers with Konnect.
+	//
+	// +optional
+	ConsumersSync *ControlPlaneKonnectConsumersSync `json:"consumersSync,omitempty"`
```
randmonkey, line 474: Are we going to configure more options on syncing consumers with Konnect? If not, I think just `enabled/disabled` for `consumerSync` is enough.

tao12345666333:
@randmonkey Do you mean https://github.com/Kong/kubernetes-ingress-controller/pull/7488？
Since it was newly added in 3.5 and did not appear in the original issue, I missed it. Let me add it.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -457,3 +462,81 @@ func (c *ControlPlane) SetConditions(conditions []metav1.Condition) {
...
+
+	// InitialPollingPeriod is the initial polling period for license checks.
+	//
+	// +optional
+	InitialPollingPeriod *metav1.Duration `json:"initialPollingPeriod,omitempty"`
+
+	// PollingPeriod is the polling period for license checks.
+	//
+	// +optional
+	PollingPeriod *metav1.Duration `json:"pollingPeriod,omitempty"`
```
pmalek, line 522: Shouldn't we add CEL rules to require these to be set when Konnect license functionality is enabled?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -457,3 +462,81 @@ func (c *ControlPlane) SetConditions(conditions []metav1.Condition) {
...
+
+	// StorageEnabled indicates whether to store licenses fetched from Konnect
+	// to Secrets locally to use them later when connection to Konnect is broken.
+	// Only effective when Enabled is set to enabled.
+	//
+	// +optional
+	// +kubebuilder:default=enabled
+	// +kubebuilder:validation:Enum=enabled;disabled
+	StorageEnabled *ControlPlaneKonnectLicensingState `json:"storageEnabled,omitempty"`
```
pmalek, line 531: Ditto about stuttering: `storageEnabled: enabled`

pmalek (changes requested):

api/gateway-operator/v2alpha1/controlplane_types.go
tao12345666333: Added here https://github.com/Kong/kubernetes-configuration/pull/535/commits/3b7bf3db2fc31d50cd8b018edc02982a788e616c

tao12345666333: However, for the current rules I added, I think they may need to be modified. These two fields have default values, so they are not necessary. What do you think?

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -457,3 +462,84 @@ func (c *ControlPlane) SetConditions(conditions []metav1.Condition) {
...
+// +kubebuilder:validation:XValidation:message="initialPollingPeriod is required when licensing is enabled",rule="!has(self.enabled) || self.enabled != 'enabled' || has(self.initialPollingPeriod)"
+// +kubebuilder:validation:XValidation:message="pollingPeriod is required when licensing is enabled",rule="!has(self.enabled) || self.enabled != 'enabled' || has(self.pollingPeriod)"
+// +kubebuilder:validation:XValidation:message="storageEnabled can only be set to enabled when licensing is enabled",rule="!has(self.storageEnabled) || self.storageEnabled == 'disabled' || self.enabled == 'enabled'"
+type ControlPlaneKonnectLicensing struct {
+	// Enabled indicates whether Konnect licensing is enabled.
+	//
+	// +optional
+	// +kubebuilder:default=disabled
+	// +kubebuilder:validation:Enum=enabled;disabled
+	Enabled *ControlPlaneKonnectLicensingState `json:"enabled,omitempty"`
```
pmalek, line 515: (Sorry I'm posting this just now, GitHub decided to remove my comment somehow during review? 😕 )

This stutters (when enabled): `enabled: enabled` or is confusing (when disabled): `enabled: disabled`.

How about changing the name of this field to `state`? I'm open to proposals on this.

api/gateway-operator/v2alpha1/controlplane_types.go
```diff
@@ -457,3 +462,84 @@ func (c *ControlPlane) SetConditions(conditions []metav1.Condition) {
...
+	// ControlPlaneKonnectConsumersSyncStateDisabled indicates that consumer synchronization is disabled.
+	ControlPlaneKonnectConsumersSyncStateDisabled ControlPlaneKonnectConsumersSyncState = "disabled"
+)
+
+// ControlPlaneKonnectLicensing defines the configuration for Konnect licensing.
+//
+// +apireference:kgo:include
+// +kubebuilder:validation:XValidation:message="initialPollingPeriod is required when licensing is enabled",rule="!has(self.enabled) || self.enabled != 'enabled' || has(self.initialPollingPeriod)"
+// +kubebuilder:validation:XValidation:message="pollingPeriod is required when licensing is enabled",rule="!has(self.enabled) || self.enabled != 'enabled' || has(self.pollingPeriod)"
+// +kubebuilder:validation:XValidation:message="storageEnabled can only be set to enabled when licensing is enabled",rule="!has(self.storageEnabled) || self.storageEnabled == 'disabled' || self.enabled == 'enabled'"
```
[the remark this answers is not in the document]
tao12345666333: There is no mandatory enable storage here; instead, according to the naming in KIC, this field is called `storageEnabled`, and it can be set to enabled or disabled.

Regarding the CEL rule, its specific action is:

1. !has(self.storageEnabled): If storageEnabled isn't specified in the configuration, validation passes.
2. self.storageEnabled == 'disabled': If storageEnabled is set to "disabled", validation passes.
3. self.enabled == 'enabled': If enabled is set to "enabled", validation passes regardless of storageEnabled's value. For storageEnabled, we have also set its optional values to be enabled or disabled.

 Do you think this rule needs to be modified?

api/gateway-operator/v2alpha1/controlplane_types.go
tao12345666333: Don't worry.
This is indeed a good point. @pmalek 

I checked the relevant settings in this repo, and now we mainly have two types of fields in this scenario, which are not unified.

* `Enabled`  can be set to true or false.
* `State` can be set to enabled or disabled.


Since this value is set to `*-enabled` in KIC, I tend to keep it consistent here by choosing enabled and setting its value to true or false. What do you think?

```
| --konnect-licensing-enabled | bool | Retrieve licenses from Konnect if available. Overrides licenses provided via the environment. | false |
```

pmalek: kube-api-linter which we currently use has a [`nobools`](https://github.com/kubernetes-sigs/kube-api-linter/blob/47bfeef6cd38540083f656ac485793fb143ddbdd/docs/linters.md#nobools) linter (which we also use) which catches usage of bools in the API (because that doesn't evolve well.

I'd suggest that being consistent here is not really required, based on the fact that we're not limited by CLI flags names but by fields structure and nesting.

`state` I **think** would fit well here (even if we've used a different name for the flag in KIC).

tao12345666333: Make sense. Let's use state.

api/gateway-operator/v2alpha1/controlplane_types.go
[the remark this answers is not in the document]
randmonkey: Sorry, I misunderstood that.

api/gateway-operator/v2alpha1/controlplane_types.go
tao12345666333: Changed. https://github.com/Kong/kubernetes-configuration/pull/535/commits/665ef8ba43e570c01423039fe8c21c6ae31a7982

pmalek (approved):

PR #536, opened 2025-07-29
chore(refactor): decouple to allow import of `v2alpha1` in `v1beta1`

programmer04:

To implement the conversion webhook, additional methods on a type must be implemented; see [WIP #488](https://github.com/Kong/kubernetes-configuration/pull/488/commits/747b20e696ecb7a5c443fd551b3769f258b7b823), which utilizes changes from this PR. Import of `v2alpha1` in `v1beta1` is required, and Go doesn't allow import cycles, so the location of particular structs has to be adjusted.

**Which issue this PR fixes**

Required for

- https://github.com/Kong/kong-operator/issues/1368

Change in `scripts/apidocs-gen/generate.sh` was dictated by a warning which appeared in logs during generation (it doesn't change the behavior in this particular change).

### DIFF
--- a/config/crd/gateway-operator/gateway-operator.konghq.com_gatewayconfigurations.yaml
+++ b/config/crd/gateway-operator/gateway-operator.konghq.com_gatewayconfigurations.yaml
@@ -26230,12 +26230,8 @@ spec:
                                             type: string
                                         type: object
                                     type: object
-                                required:
-                                - promotion
                                 type: object
                             type: object
-                        required:
-                        - strategy
                         type: object
                       scaling:
                         description: Scaling defines the scaling options for the deployment.


api/gateway-operator/v2alpha1/gatewayconfiguration_types.go
```diff
@@ -93,7 +93,7 @@ type GatewayConfigControlPlaneOptions struct {
 // +apireference:kgo:include
 type GatewayConfigDataPlaneOptions struct {
 	// +optional
-	Deployment operatorv1beta1.DataPlaneDeploymentOptions `json:"deployment"`
```
pmalek, line 96: Wouldn't it be easier to just redefine `DataPlaneDeploymentOptions` in `v2alpha1` pkg and use that 🤔 ?

api/common/v1beta1/deployment_options.go
pmalek, line 1: Wouldn't we want to copy this into v2alpha1 as well?

This would allow us to decouple `DeploymentOptions` and watch namespace options. WDYT?

programmer04: so I coiped all required, pelase double check 4cd02a5ac88d0e29b18568eedac5692e67164158 that was enforced by linter

config/crd/gateway-operator/gateway-operator.konghq.com_gatewayconfigurations.yaml
pmalek, line 1: Seems like these were incorrectly set to required having defaults. Looks like kube-api-linter proved itself useful to detect this.

pmalek (approved):

PR #550, opened 2025-08-07
feat(conversion): implement v1alpha1 <-> v1alpha2 conversion for KonnectGatewayControlPlane

alacuku:

- Refactored v1alpha1 to depend on v1alpha2 for Konnect API types to break import cycles and enable implementation of webhook conversion methods between versions.
- Add ConvertTo and ConvertFrom methods for KonnectGatewayControlPlane to enable conversion between v1alpha1 and v1alpha2 API versions;
- Add unit tests for the conversion functions to verify correct behavior;

**Which issue this PR fixes**

Fixes #

**PR Readiness Checklist**:


.github/workflows/tests.yaml
```diff
@@ -87,6 +87,22 @@ jobs:
     - name: Run tests
       run: make test.unit.pretty
 
+  test-conversion:
```
pmalek, line 90: 👍 

Just a nit: we need to add it to `passed` job's `needs` so that we actually require it to pass for a PR to be allowed to merge.

This list is defined here: https://github.com/Kong/kubernetes-configuration/blob/a3bbc4e797977ce17c324a8665be651962b70055/.github/workflows/tests.yaml#L275-L286

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
```diff
@@ -196,7 +197,7 @@ type ConfigurationDataPlaneGroupAutoscaleStatic struct {
 // KonnectCloudGatewayDataPlaneGroupConfigurationStatus defines the observed state of KonnectCloudGatewayDataPlaneGroupConfiguration.
 // +apireference:kgo:include
 type KonnectCloudGatewayDataPlaneGroupConfigurationStatus struct {
-	KonnectEntityStatusWithControlPlaneRef `json:",inline"`
+	konnectv1alpha2.KonnectEntityStatusWithControlPlaneRef `json:",inline"`
```
pmalek, line 200: Couldn't we just leave the `KonnectEntityStatusWithControlPlaneRef` defined in the `v1alpha1` package here and use it? 🤔 

This way we wouldn't have to import `github.com/kong/kubernetes-configuration/v2/api/konnect/v1alpha2`.

Same applies elsewhere.

api/konnect/v1alpha1/konnect_gateway_controlplane_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+	if src.Spec.Mirror != nil {
+		dst.Spec.Mirror = &konnectv1alpha2.MirrorSpec{
+			Konnect: konnectv1alpha2.MirrorKonnect{
+				ID: src.Spec.Mirror.Konnect.ID,
+			},
+		}
+	} else {
+		dst.Spec.Mirror = nil
+	}
+	dst.Spec.Source = src.Spec.Source
```
randmonkey, line 30: Should we use `DeepCopy` method or other dedicated ways here for pointer values and slices?

api/konnect/v1alpha1/konnect_gateway_controlplane_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+// handling nil pointers and type conversions appropriately.
+func createControlplaneRequestFromSpec(spec KonnectGatewayControlPlaneSpec) *sdkkonnectcomp.CreateControlPlaneRequest {
+	// Only create the request if this is an Origin type (not Mirror)
+	if spec.Source != nil && *spec.Source == "Mirror" {
+		return nil
+	}
+
+	return &sdkkonnectcomp.CreateControlPlaneRequest{
+		Name:         lo.FromPtr(spec.Name),
+		Description:  spec.Description,
```
randmonkey, line 72: Should we also deepcopy the pointer values, slices and maps here?

api/konnect/v1alpha1/konnect_gateway_controlplane_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+		Labels:       spec.Labels,
+	}
+}
+
+// createInlineControlPlaneRequest fills a v1alpha1 CreateControlPlaneRequest from an sdk CreateControlPlaneRequest.
+func createInlineControlPlaneRequest(req *sdkkonnectcomp.CreateControlPlaneRequest) CreateControlPlaneRequest {
+	if req == nil {
+		return CreateControlPlaneRequest{}
+	}
+	return CreateControlPlaneRequest{
```
randmonkey, line 86: Ditto here.

.github/workflows/tests.yaml
alacuku: Nice catch!

api/konnect/v1alpha1/konnect_gateway_controlplane_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	sdkkonnectcomp "github.com/Kong/sdk-konnect-go/models/components"
+
+	konnectv1alpha2 "github.com/kong/kubernetes-configuration/v2/api/konnect/v1alpha2"
+)
+
+// ConvertTo converts this KonnectGatewayControlPlane (v1alpha1) to the Hub version (v1alpha2).
+func (src *KonnectGatewayControlPlane) ConvertTo(dstRaw conversion.Hub) error {
```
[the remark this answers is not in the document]
alacuku: We assume the controller-runtime machinery always provides valid, non-nil objects. If that happens it's a bug in the controller runtime then.

Maybe we can check that we were able to convert the arguments to the right type.

api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
pmalek: Ok, I see.

Even though I do believe that "a little copying is better than little dependency" [link](https://go-proverbs.github.io/) [link_yt](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s) I do see that this would require more work now for little benefit (given that `v1alpha1` is going away and that Go is not our primary interface for end users).

I'm OK leaving this as is then. 👍

mlavacca (approved):
This lgtm. Once other threads are resolved, it's good to be merged from my point of view

api/konnect/v1alpha1/konnect_gateway_controlplane_conversion.go
alacuku: None of this fields have a dedicated `DeepCopy` method.

As far as I can tell, it's safe to assume that the `source` object is destroyed at the end of the conversion, making the `destination` object the owner of all references assigned to it. We expect that the pointer/complex structs assignments are safe to do without the need to `deepCopy` them. 
https://github.com/kubernetes-sigs/controller-runtime/blob/c20ea143a236a34fb331e6c04820b75aac444e7d/pkg/webhook/conversion/conversion.go#L90-L149

PR #555, opened 2025-08-11
feat(controlplane): add dataplane in status and enforce ingressClass for ref type dataplane targets

pmalek:

This PR aims to support the 2 use cases for the `ControlPlane` API:

- user created `ControlPlane` (together with a `DataPlane`), this will require `spec.ingressClass` to be set and will only support the Ingress reconciliation (as there's no `Gateway` to idiomatically support Gateway API).
- controller created `ControlPlane`:
  - this use case will support both Ingress (when `spec.controlPlaneOptions.ingressClass` is set in `GatewayConfiguration` by the user) and the Gateway API

The above mentioned is enforced by the CRD CEL validation rules.

Additionally, to save the information about the target `DataPlane`, `status.dataplane` field is added.

**Which issue this PR fixes**

Part of https://github.com/Kong/kong-operator/issues/1358

**PR Readiness Checklist**:

### DIFF
--- a/api/gateway-operator/controlplane/conditions.go
+++ b/api/gateway-operator/controlplane/conditions.go
@@ -8,8 +8,7 @@ import "github.com/kong/kubernetes-configuration/v2/api/common/consts"
 
 const (
 	// ConditionTypeProvisioned is a condition type indicating whether or
-	// not all Deployments (or Daemonsets) for the ControlPlane have been provisioned
-	// successfully.
+	// not the ControlPlane has been provisioned successfully.
 	ConditionTypeProvisioned consts.ConditionType = "Provisioned"
 
 	// ConditionTypeWatchNamespaceGrantValid is a condition type used to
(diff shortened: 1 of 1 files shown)


api/gateway-operator/v2beta1/controlplane_types.go
```diff
@@ -65,6 +65,7 @@ type ControlPlaneList struct {
 // ControlPlaneSpec defines the desired state of ControlPlane
 //
 // +apireference:kgo:include
+// +kubebuilder:validation:XValidation:message="",rule="self.dataplane.type != 'ref' || has(self.ingressClass)"
```
mlavacca, line 68:
```suggestion
// +kubebuilder:validation:XValidation:message="",rule="self.dataplane.type != 'ref' && has(self.ingressClass)"
```

Shouldn't this be `&&`?
Also, can we add a message here?

mlavacca (commented):
As we spoke offline, can we add a condition related to the resource being `managed-by-owner` but without the `Gateway` `OwnerRef`?

api/gateway-operator/v2beta1/controlplane_types.go
pmalek: If the type is set to something else than `ref` then the condition exits early (not checking the `self.ingressClass`.

If it's `ref` (then the first part does not equal to `true`) then we check the second part and enforce the ingress class being set.

---

Adding the message in 94d73c6e6f9e0e89271775b2b49c64b005b6cc42.

pmalek:
> As we spoke offline, can we add a condition related to the resource being `managed-by-owner` but without the `Gateway` `OwnerRef`?

👍 Adding the condition reason for the `Provisioned` condition in 3925b3994377081a0d61ecabe932e238fb888ce7.

mlavacca (approved):

PR #580, opened 2025-08-25
chore: deduplicate KIC and KGO CRDs and remove deprecated `gateway-operator.konghq.com` `KonnectExtension`

programmer04:

- Two sets of CRDs `config/crd/gateway-operator` and `config/crd/ingress-controller` have some resources duplicated. It leads to problems when they are used with Helm as proposed in https://github.com/Kong/kong-operator/pull/2122 - Helm ownership metadata reports a conflict like that 
     ```log
     ...exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal...
     ```
     So moving it only to be in KGO for now resolves the problem

- `KonnectExtension` in the group `gateway-operator.konghq.com` is deprecated in favor of `KonnectExtension` `konnect.konghq.com` since a couple of releases. There is no way to deal with conversion correctly in this case, so it's removed entirely. In such a case, it will fail instead of being accepted (and not doing anything).

**Which issue this PR fixes**

Required by https://github.com/Kong/kong-operator/pull/2122 which resolves https://github.com/Kong/kong-operator/issues/1987

It's just to unblock for now. Ideally, both CRD sets should be merged into one KO set. It is something that should be addressed in a separate issue, probably together with https://github.com/Kong/kong-operator/issues/1960

**PR Readiness Checklist**:


pmalek (approved):
I believe we're safe to make this change.

These CRDs are still used by KIC and by KIC's chart (https://github.com/Kong/charts/blob/0fd1503b2f5836d5e340df5a20cca250cab0b017/charts/kong/crds/custom-resource-definitions.yaml) but that's limited to (currently) 1.5.x release branch of kcfg + we're not planning to release further minor/major releases of KIC.

PR #593, opened 2025-08-28
Deprecate static autoscale in KonnectCloudGatewayDataPlaneConfiguration

fgiudici:

**Which issue this PR fixes**

Related to https://github.com/Kong/kong-operator/issues/1779

**PR Readiness Checklist**:

### DIFF
--- a/api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
+++ b/api/konnect/v1alpha1/konnect_cloudgateway_dataplane_configuration_types.go
@@ -136,6 +136,9 @@ type ConfigurationDataPlaneGroupAutoscaleType string
 
 const (
 	// ConfigurationDataPlaneGroupAutoscaleTypeStatic is the autoscale type for static configuration.
+	//
+	// Deprecated: ConfigurationDataPlaneGroupAutoscaleTypeStatic is deprecated. Use
+	// ConfigurationDataPlaneGroupAutoscaleTypeAutopilot instead.
 	ConfigurationDataPlaneGroupAutoscaleTypeStatic ConfigurationDataPlaneGroupAutoscaleType = "static"
 
 	// ConfigurationDataPlaneGroupAutoscaleTypeAutopilot is the autoscale type for autopilot configuration.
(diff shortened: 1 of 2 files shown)


fgiudici:
ConfigurationDataPlaneGroupAutoscaleTypeStatic is  the "static" value for the `autoscale` property in the `KonnectCloudGatewayDataPlaneGroupConfiguration` resources. It maps to the `sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleStatic` value, which is deprecated in the Konnect API.
Marking our `ConfigurationDataPlaneGroupAutoscaleTypeStatic` constant as deprecated means that it should not be used in code anymore. Still, what we really want is that the Konnect API is not used with that value.
For this reason I am leaning to not mark the constant deprecated: we will manage the deprecation with a warning when the Kubernetes resources is used with that value using a ValidationAdmissionPolicy (see https://github.com/Kong/kong-operator/pull/2153).
Closing this PR.

pmalek:
> ConfigurationDataPlaneGroupAutoscaleTypeStatic is the "static" value for the `autoscale` property in the `KonnectCloudGatewayDataPlaneGroupConfiguration` resources. It maps to the `sdkkonnectcomp.ConfigurationDataPlaneGroupAutoscaleStatic` value, which is deprecated in the Konnect API. Marking our `ConfigurationDataPlaneGroupAutoscaleTypeStatic` constant as deprecated means that it should not be used in code anymore. Still, what we really want is that the Konnect API is not used with that value. For this reason I am leaning to not mark the constant deprecated: we will manage the deprecation with a warning when the Kubernetes resources is used with that value using a ValidationAdmissionPolicy (see [Kong/kong-operator#2153](https://github.com/Kong/kong-operator/pull/2153)). Closing this PR.

IMHO we want both. Kubernetes API restriction (using `ValidationAdmissionPolicy`) will indeed help us here but it's later on in the chain: ideally, users who use the Go API would get the deprecation warning (we can't deny this usage at the Go API level, even with removing the `const`) earlier than with the `ValidationAdmissionPolicy`.

I agree it's not strictly necessary but it would be a nice touch. Just my 2 cents.

